PR #268, opened 2026-03-29
Cap sidebar nav badges at 99+ for large counts

canalesb93:
## Summary
- Nav badges (Connections, Review Queue, Reports) now show "99+" instead of raw counts like "555"
- Added `title` tooltip with exact count on hover
- New `badgeCount` template function


canalesb93:
Don't want this change.

PR #289, opened 2026-03-30
Fix sync engine re-applying rules to unchanged transactions

canalesb93:
## Summary
- Split `upsertTransaction` into two steps: pure upsert + conditional `applyRulesToTransaction`, so rules are only evaluated for new or changed transactions
- Fixed `classifyUpsertResult` to check `created_at` recency, preventing recently-added rows from being misclassified as "new" on re-sync
- Moved rule application inserts and category updates inside the DB transaction (previously used pool outside tx)
- Added `WHERE action_value IS DISTINCT FROM` guard on rule application upsert to prevent no-op timestamp bumps

## Test plan
- [x] New integration test `TestSync_UnchangedTransactions_SkipRuleReapplication` verifies unchanged transactions skip rule evaluation, don't bump `applied_at`, don't increment `hit_count`, and show correct sync log counts
- [x] New unit test case for `classifyUpsertResult` covering re-sync of recently added rows
- [x] All 22 existing sync integration tests pass
- [x] All unit tests pass

### DIFF
--- a/internal/sync/engine.go
+++ b/internal/sync/engine.go
@@ -323,12 +323,12 @@ func (e *Engine) runSync(ctx context.Context, connectionID pgtype.UUID, logger *
 				skipped++
 				continue
 			}
-			ur, err := e.upsertTransaction(ctx, txQueries, &pendingAdded[i], accountIDCache, string(conn.Provider), resolver, conn.UserID, userName, logger)
+			dbTxn, err := e.upsertTransaction(ctx, txQueries, &pendingAdded[i], accountIDCache)
 			if err != nil {
 				logger.Error("upsert added transaction", "external_id", pendingAdded[i].ExternalID, "error", err)
 				continue
 			}
-			isNew, isChanged := classifyUpsertResult(ur.Txn, upsertStart)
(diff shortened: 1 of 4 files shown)


canalesb93:
## Future alternatives considered

**Replace timestamp heuristic with deterministic INSERT/UPDATE signal**: The `classifyUpsertResult` function uses a timing-based heuristic (`created_at` vs `upsertStart` with 2-second tolerance) to distinguish new rows from unchanged re-syncs. This works in production (syncs are minutes apart) but forces a 3-second sleep in the integration test. A cleaner approach would be to have the upsert SQL return an explicit signal — e.g., PostgreSQL's `xmax = 0` trick or a `CASE WHEN xmax = 0 THEN true ELSE false END AS is_insert` column in the RETURNING clause. This would eliminate the timing heuristic entirely but requires modifying the sqlc query and regenerating. Worth doing if the heuristic causes friction elsewhere.

**Minor redundancy**: `applyRulesToTransaction` re-resolves `accountID` via `resolveAccountID` even though the caller already has it. Could pass it as a parameter to avoid the extra cache lookup.

Issue #433, opened 2026-04-17
Remaining audit §D test additions not yet landed


canalesb93:
## Context

The audit doc `docs/planned-features/transaction-rules-test-gaps.md` §D listed a fairly long set of test additions for the rule system. The polish project (Phases 0–5) landed most of the critical ones, but a handful of lower-priority tests from that list never got added. Tracking them here so they don't get lost.

### Integration (engine_integration_test.go)

- `TestRule_ExpiredRule_NotFired` — a rule whose `expires_at` is in the past must be excluded from sync.
- `TestRule_DisabledRule_NotFired_DuringSync` — a disabled rule must be excluded from sync (`ListActiveRulesForSync` filter is tested at list level but not sync level end-to-end).
- `TestRule_OverrideSuppressesSetCategoryButNotOthers` — a rule with `[set_category, add_tag]` applied to a `category_override=TRUE` transaction should skip set_category but still add the tag. Partially covered by resolver-level unit tests; integration case would verify DB state.
- `TestRule_HitCountMatchesWrites_TagOnly` — rule with only `add_tag` against already-tagged txn; hit_count still bumps (= "matched" semantic; pins Q12/Q16 interaction).
- `TestRule_AllActionsAtomicInSyncTx` — force failure after set_category UPDATE; verify full rollback.

### Retroactive (service integration tests)

- `TestApplyRuleRetroactively_Expired_Rejected` — expired rule can't be applied retroactively.
- `TestApplyRuleRetroactively_Disabled_Rejected` — disabled rule can't be applied retroactively.
- `TestApplyAllRulesRetroactively_SamePriorityTie` — same-priority tiebreak in retroactive path should match sync's `created_at ASC` rule.
- `TestPreviewRule_DoesNotModify` — scan annotations/categories counts before + after preview; assert zero delta.
- `TestPreviewRuleForDetail_ExcludesAlreadyApplied` — the `excludeRuleID` branch that skips transactions with a prior rule_applied annotation for this rule.

### Unit (rule_resolver_test.go)

- `TestResolveWithContext_AddCommentAccumulation_Dedup` — two rules adding identical comment content; verify both land (no dedup, by design) per audit doc §C concern.
- `TestResolveWithContext_RuleWithOnlyTagAction_NoCategoryWrite` — a tag-only rule should not touch transactions.category_id even when a set_category rule would have applied elsewhere.

## Why they were deferred

[...]

PR #440, opened 2026-04-17
chore(rules): small rule-engine code-quality cleanups

### LINKED ISSUE #434: Small rule-engine code-quality cleanups (dead code, silent failures, duplication)
## Context

Grab-bag of small improvements around the rule engine — each is low-risk and independent. Batched here so they don't each need their own issue.

### Dead code: `buildActionSetClause`

`internal/service/rules.go buildActionSetClause` (~line 2119) is defined but no longer called after the Phase 2 refactor extracted action intents inline in `ApplyRuleRetroactively`. Can be deleted.

### Silent failure in `loadTagSlugsInTx`

`internal/sync/engine.go loadTagSlugsInTx` returns `(slugs, err)`, and the caller does `if err == nil { tctx.Tags = slugs }` — an error silently results in empty Tags with no log. Small log line on the error path would help debug future tag-loading issues without changing behavior.

### Duplicated `titleCaseSlug` / `titleCaseSlugForRule`

`internal/sync/engine.go titleCaseSlugForRule` and `internal/service/tags.go titleCaseSlug` are near-identical. The duplication was intentional (sync package shouldn't import service), but a shared helper in a neutral package (e.g. `internal/slugs/`) would collapse them.

### `rule_applied` annotation payload size

When a rule's action is `add_comment`, the full comment content lands in the `rule_applied` annotation's `action_value` field **and** in the separate `comment` annotation. For long comments this doubles the DB footprint. Consider storing just a content digest or a truncation marker in `rule_applied`.

### `sourceByKey` first-rule-wins on duplicate tag slugs

If two rules both add tag "foo", `sourceByKey["tag|foo"]` keeps only the first rule's metadata. The `transaction_tags.added_by_*` columns credit only that first rule. This matches the `ON CONFLICT DO NOTHING` persistence but loses a bit of provenance. Either record the multiple contributors in the annotation payload or document the first-wins behavior.

### `UpdateTransactionRule` unreachable branch

`internal/service/rules.go` has:
```go
trigger := existing.Trigger
if trigger == "" { trigger = "on_create" }
```
The DB column is NOT NULL DEFAULT 'on_create', so `existing.Trigger == ""` is unreachable. Minor — the defensive code is cheap but misleading.

### `TestResolveWithContext_PriorityOrdering_LastWins` tests in-slice order, not DB ordering

The unit test passes rules as a pre-ordered slice and asserts the later-in-slice one wins. That's a test of the resolver's last-wins merge, but it doesn't exercise the `ORDER BY priority ASC, created_at ASC` load at the DB level. A focused integration test would close the loop (see also #433).

## Why these were deferred

Each is cosmetic or minor — not blocking any functional behavior. Grouped here so they can be fixed in one cleanup PR or individually picked off as good-first-issue work.

## Context links

- `docs/planned-features/transaction-rules-test-gaps.md` §C — most of these items come from the audit's "suspicious / surprising code" list

canalesb93:
## Summary
- Delete `buildActionSetClause` in `internal/service/rules.go` — dead after Phase 2.
- Log (instead of silently dropping) errors from `loadTagSlugsInTx` in `internal/sync/engine.go`. Behavior unchanged otherwise.
- Extract `titleCaseSlug` / `titleCaseSlugForRule` into a neutral `internal/slugs` package so `internal/service` and `internal/sync` share one implementation.
- Remove unreachable `if trigger == ""` branch from `UpdateTransactionRule` (column is `NOT NULL DEFAULT`, `normalizeTrigger` guarantees non-empty).
- Document the first-wins semantic for duplicate `(field, value)` keys in `sourceByKey` so the comment matches observable behavior.

Skipped from the issue:
- `rule_applied` payload-size cleanup (non-trivial schema change; left as follow-up, per issue notes).
- Integration-test suggestion (covered separately by #433).

Closes #434

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./...`
- [x] `make test-integration`

### DIFF
--- a/internal/service/rules.go
+++ b/internal/service/rules.go
@@ -9,6 +9,8 @@ import (
 	"strings"
 	"time"
 
+	"breadbox/internal/slugs"
+
 	"github.com/jackc/pgx/v5"
 	"github.com/jackc/pgx/v5/pgtype"
 )
@@ -986,6 +988,9 @@ func (s *Service) UpdateTransactionRule(ctx context.Context, id string, params U
 		actions = newActions
 	}
 
+	// existing.Trigger is sourced from a NOT NULL DEFAULT column, and
+	// normalizeTrigger guarantees a non-empty value on update — no fallback
+	// needed here.
 	trigger := existing.Trigger
 	if params.Trigger != nil {
 		t, err := normalizeTrigger(*params.Trigger)
@@ -993,9 +998,6 @@ func (s *Service) UpdateTransactionRule(ctx context.Context, id string, params U
 			return nil, err
 		}
 		trigger = t
-	}
-	if trigger == "" {
-		trigger = "on_create"
 	}
 
 	priority := int32(existing.Priority)
@@ -2064,7 +2066,7 @@ func (s *Service) materializeRuleTagAdd(ctx context.Context, tx pgx.Tx, txnID pg
 			INSERT INTO tags (slug, display_name, lifecycle)
 			VALUES ($1, $2, 'persistent')
 			ON CONFLICT (slug) DO UPDATE SET updated_at = tags.updated_at
-			RETURNING id`, slug, titleCaseSlug(slug)).Scan(&tagID); err2 != nil {
+			RETURNING id`, slug, slugs.TitleCase(slug)).Scan(&tagID); err2 != nil {
 			return false, fmt.Errorf("get or create tag %q: %w", slug, err2)
 		}
 	}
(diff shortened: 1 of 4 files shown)


canalesb93:
Closing as redundant — all 5 cleanups from issue #434 have already landed on `main` via PRs #442 and #444:

1. `buildActionSetClause` — no longer present in `internal/service/rules.go` on `main`.
2. Log line on error path in `loadTagSlugsInTx` — present on `main` (uses `e.logger.Warn("load tag slugs for rule evaluation failed; continuing without tag context", ...)`).
3. `titleCaseSlug`/`titleCaseSlugForRule` extraction to `internal/slugs/` — `internal/slugs/slugs.go` exists on `main`, consumed by `internal/sync/engine.go`, `internal/service/rules.go`, and `internal/service/tags.go` via `slugs.TitleCase(...)`.
4. Unreachable `if trigger == ""` branch in `UpdateTransactionRule` — already removed on `main`; replaced with a clarifying comment about `normalizeTrigger` guaranteeing a non-empty value.
5. First-rule-wins comment for `sourceByKey` — present on `main` verbatim.

After rebasing this branch onto `origin/main`, the commit became empty and git dropped it. No follow-up work is needed — issue #434 is effectively resolved.

Issue #435, opened 2026-04-17
Unify transaction display: extract shared row component + normalize tx preview DTO


canalesb93:
## Background

PR #429 polished how transactions render in the **global command palette** to match the main transactions list. While doing that I inventoried every surface that displays a transaction row/card, and there is meaningful duplication worth consolidating. Filing this as a tracking issue so the work can happen incrementally.

Context trail:
- PR #429 — polish cmdk transaction results (merged)
- `internal/templates/partials/tx_row.html` — canonical heavy row used by the main list
- `internal/admin/transactions.go:1131-1200` — `QuickSearchTransactionsHandler` (JSON for cmdk)
- `internal/admin/rules.go:464` — `PreviewRuleAdminHandler` (JSON for rule preview modal)

### Server-rendered, already share `tx-row`
- `pages/transactions.html` + `partials/tx_results.html` (main list, grouped + flat)
- `pages/account_detail.html` (account transactions tab) — lines ~135-160

### Server-rendered, **bespoke markup** (candidates for consolidation)
- `pages/dashboard.html:254-304` — recent transactions widget (avatar pill + inline text)
- `pages/rule_detail.html:224-259` — "Recent Applications" table rows
- `pages/rule_detail.html:282-307` — "Uncategorized Matches" table rows
- `pages/account_link_detail.html:26-82` — match-pair table (different enough to leave alone; see non-goals)

### Client-rendered from JSON (needs shared JS/CSS)
- **Command palette** — `/-/search/transactions` endpoint; JSON shape was just expanded in #429 (`amount_label`, `date_label`, `user_name`, `category_display_name`, `category_icon`, `category_color`). Alpine renders it inline in `internal/templates/layout/base.html` using `.bb-cmdk-tx-*` classes.
- **Rule preview modal** — `/-/rules/preview` endpoint returning `RulePreviewResult.SampleMatches`. Currently ships `{name, amount, date, category_slug}` — no icon, no color, no user, no account. Rendered client-side by the rule form; visual style drifts from the palette.

### Layer 1 — Extract `tx-row-compact` partial (Go template)

Lightweight sibling of `tx-row`: avatar, name + Pending badge, `user · account · category (dot)` meta line, date + amount on the right. No checkbox, no inline category picker, no expand panel.

Input: existing `service.AdminTransactionRow` (no new DTO).

**Adopters:**
- `pages/dashboard.html` recent transactions
- `pages/rule_detail.html` recent applications table
[...]

PR #447, opened 2026-04-17
refactor(ui): migrate rule_detail.html tables to tx-row-compact

canalesb93:
## Summary

PR 2 of 3 for #435 (Partial fix for #435 — do not close).

Migrates both tables on `/rules/<id>` to the shared `tx-row-compact` partial from PR 1 (#443), so rule detail matches the visual density of the dashboard's Recent Transactions feed.

- **Recent Applications** — compact row per application plus a thin secondary meta line underneath showing the action (Set category / Tag / Comment) and how it fired (`sync` / `retro` / `manual`). The five-column table is gone.
- **Matching transactions** (preview) — compact rows only. The rule's action category is already in the section header (`Categorize as ...`), and the current category is rendered by the compact row's category pill, so the `Current Category` column is redundant. Headers dropped on both sections for cleaner visual polish — judgment call per the brief.

Adds `Service.GetAdminTransactionRowsByIDs` so the rule detail handler can hydrate `RuleApplicationRow` / `RulePreviewMatch` IDs into the full `AdminTransactionRow` shape the partial consumes. Preserves input order. A richer `TransactionSummary` DTO with a reusable batch loader lives in PR 3 scope.

### Recent Applications

<table>
  <tr>
    <td>Recent Applications — before</td>
    <td>Recent Applications — after</td>
  </tr>
</table>

### Matching transactions (preview)

<table>
  <tr>
    <td>Matching transactions — before</td>
    <td>Matching transactions — after</td>
  </tr>
</table>

## Test plan

- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `go test ./...` passes
- [x] `make css` rebuild (no new utilities introduced, ran for safety)
- [x] Visually verified at 1280x900 on `/rules/<id>` for a rule with recent applications (`General -> Needs Enrichment`) and a separate rule with untouched matches (`PR2 Preview Test - Zelle`, 26 pending).

### DIFF
--- a/internal/admin/rules.go
+++ b/internal/admin/rules.go
@@ -279,24 +279,65 @@ func RuleDetailPageHandler(svc *service.Service, sm *scs.SessionManager, tr *Tem
 		// Recent applications
 		applications, hasMoreApps, _ := svc.ListRuleApplications(ctx, id, 10, "")
 
+		// Category tree powers (a) the inline category picker on tx-row, and
+		// (b) slug → display-name resolution for the action meta line shown
+		// above each Recent Applications row.
+		categories, _ := svc.ListCategoryTree(ctx)
+		type catMeta struct {
+			DisplayName string
+			Color       *string
+			Icon        *string
+		}
+		catBySlug := make(map[string]catMeta)
+		for _, p := range categories {
+			catBySlug[p.Slug] = catMeta{DisplayName: p.DisplayName, Color: p.Color, Icon: p.Icon}
+			for _, c := range p.Children {
+				catBySlug[c.Slug] = catMeta{DisplayName: c.DisplayName, Color: c.Color, Icon: c.Icon}
+			}
+		}
+
 		// Hydrate application rows with AdminTransactionRow data so the shared
-		// tx-row-compact partial can render them (category avatar, account, user,
-		// agent-reviewed flag, pending state). Preserves order.
+		// tx-row partial can render them (category avatar, account, user,
+		// agent-reviewed flag, pending state). Preserves order. ApplicationMeta
+		// carries the per-application action so the rule_detail template can
+		// prefix each row with a prominent "what happened" pill.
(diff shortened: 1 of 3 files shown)


canalesb93:
Not sure if these should use tx-row-compact, why not use regular rows.

Also we need to handle better the Action / How columns in the rules screen. In the new design it's less clear what happened.

canalesb93:
Addressed the feedback:

- **Swapped `tx-row-compact` for the regular `tx-row`** on both sections. The rule detail now matches the density and feature-set of the main transactions list (inline category picker, expand panel, rich meta line). Added the usual runtime deps (`window.__bbCategories`, `quickSetCategory`, minimal bulk-store shim) so the picker works on this page without the full transactions-page scaffolding.
- **Recent Applications now shows a prominent action strip above each row**: _"Added tag → needs-review"_, _"Set category → Groceries"_ (with the real category's color dot), _"Added comment → …"_ — plus an applied-by pill (`via sync` / `retroactive`) on the right. Category action values are resolved from slug → display name so it reads like the real category rather than the slug.
- **Dropped the `Current Category` column** on the preview table (fallback only) — `tx-row`'s inline category pill already conveys it. The preview rows use the shared partial in the common path.
- **On "how":** the `rule_applied` annotation payload doesn't persist the matched condition / field-value per-row, only `action_field`/`action_value`/`applied_by`. The rule's trigger condition is already shown at the top of the page ("What this rule does → WHEN"), so I didn't duplicate it per row — it'd be identical on every line. If you want per-row "how" later, that needs a payload-schema change on the annotation writer.

<br>

<table>
  <tr>
    <th>Top of /rules/&lt;id&gt; (action strips visible)</th>
    <th>Matching transactions (tx-row)</th>
  </tr>
  <tr>
    <td>rule top — after</td>
    <td>matching — after</td>
  </tr>
</table>

Recent Applications — full section with action/how clarity

PR #585, opened 2026-04-19
fix(connections): make sync error message touch/keyboard accessible

### LINKED ISSUE #576: mobile(connections): Sync Error tooltip is hover-only, unreadable on touch
## Route + Viewports

- Route: `/connections` (list view)
- Viewports affected: Mobile 500×844 (primary), Tablet 820×1180. Also any touch-only device at Desktop.

## Current

Connections with a failed last sync surface the problem via a small red `Sync Error` badge. The underlying error message (e.g., "Could not retrieve accounts from Teller.") is only revealed via a CSS `group-hover` tooltip — which is unreachable on touch devices and invisible to keyboard users.

list view on mobile — Sync Error pill with no affordance to see the error

On mobile, the user sees that CapitalOne and Chase both have a "Sync Error" pill, but there is no way to read the actual error without drilling into the connection detail (where, see the sibling status-badge issue, the header still says "Active"). For a self-hosted app where a household owner is usually the one diagnosing sync failures, this is the first place the problem should be legible.

## Expected

The error message should be accessible without hover:

1. On mobile, show a short error description below the connection row when `LastSyncStatus == "error"` (similar to the existing ITEM_LOGIN_REQUIRED warning banner pattern already used on the same page, lines 182–194 of `connections.html`).
2. Alternatively, the `Sync Error` pill should be a focusable button/details element that expands the message on tap/Enter.

Either way, parity with the desktop hover state on touch + keyboard is the goal.

## Repro

1. Have a connection whose most recent sync failed (`last_sync_status = 'error'`) but whose overall `status` is still `active`.
2. Resize browser to 500×844 (Chrome DevTools mobile emulation) and navigate to `/connections`.
3. Tap the "Sync Error" pill or its row — nothing reveals the message.

## Proposed fix

- **File:** `internal/templates/pages/connections.html`
- **Change:** In the `<!-- ... -->` branch that renders the `Sync Error` badge (around lines 124–133), also render a small inline error banner under the header row (inside the connection card, before `.Accounts`) when `LastSyncErrorMessage.Valid` and the current tooltip-only pattern is the only feedback. The existing `ErrorMessage.Valid` block at lines 182–194 is a good template — it already renders a `mx-4 sm:mx-6 mb-4` warning strip that works without hover. Extend that block (or add a sibling) to trigger on `and (eq .LastSyncStatus "error") (ne (printf "%s" .Status) "error")`, with `{{.LastSyncErrorMessage.String}}` as the message.
- **Rationale:** Reuses an existing, touch-accessible banner pattern already on the same page, avoids hover-only affordances, and keeps the row compact when there is no error.

## Acceptance criteria

**Before:**
- Mobile (500×844): "Sync Error" pill is visible but the actual error cannot be read from `/connections`.
- Tablet (820×1180): Same, no hover on touch.
- Desktop (1440×900): Hover tooltip works.

**After:**
- Mobile (500×844): The error message is visible below the connection header without interaction (or is reachable by tap).
- Tablet (820×1180): Same.
- Desktop (1440×900): Error is visible inline; the hover tooltip may be removed or kept, but is no longer the only way to see the message.

canalesb93:
Fixes #576

## Summary

The `Sync Error` badge on `/connections` revealed its error message only via a CSS `group-hover` tooltip — unreachable on touch devices and invisible to keyboard users. Replaced the hover-only tooltip with an inline error banner below the connection header (same pattern already used for the reauth warning on lines 182–194 of `connections.html`), so the error is legible on mobile, tablet, and desktop without any interaction.

The old tooltip (`cursor-help` span + hidden absolutely-positioned popover) was removed to avoid redundancy now that the message is always visible inline.

## Before / After

| Viewport | Before | After |
|---|---|---|
| Mobile (500×844) | before mobile | after mobile |
| Tablet (820×1180) | before tablet | after tablet |
| Desktop (1440×900) | before desktop | after desktop |

## Test plan

- [x] `go build ./...` passes
- [x] `/connections` on mobile (500×844): sync error message visible inline under the connection header without any interaction
- [x] Tablet (820×1180) and desktop (1440×900): same — message is visible inline, no hover needed
- [x] Connections with `.ErrorMessage.Valid` (reauth/login-required) still show the existing warning banner — the new inline block only renders when there's no connection-level error and `last_sync_status == "error"` with `status == "active"`
- [x] Connections with `last_sync_status == "success"` render the usual `Active` badge and no banner

### DIFF
--- a/internal/templates/pages/connections.html
+++ b/internal/templates/pages/connections.html
@@ -122,15 +122,7 @@
               {{if .IsStale}}
               <span class="text-[0.7rem] font-medium px-2 py-0.5 rounded-full bg-warning/10 text-warning">Stale</span>
               {{else if and (eq .LastSyncStatus "error") (eq (printf "%s" .Status) "active")}}
-              <span class="relative group/err">
-                <span class="text-[0.7rem] font-medium px-2 py-0.5 rounded-full bg-error/10 text-error cursor-help">Sync Error</span>
-                {{if .LastSyncErrorMessage.Valid}}
-                <span class="absolute left-0 top-full mt-1.5 z-50 hidden group-hover/err:block w-64 sm:w-80 px-3 py-2.5 rounded-lg bg-base-300 text-xs shadow-lg border border-error/20 leading-relaxed">
-                  <span class="flex items-center gap-1.5 text-error font-medium mb-1"><i data-lucide="alert-circle" class="w-3 h-3"></i>Last sync failed</span>
-                  <span class="text-base-content/70">{{.LastSyncErrorMessage.String}}</span>
-                </span>
-                {{end}}
-              </span>
+              <span class="text-[0.7rem] font-medium px-2 py-0.5 rounded-full bg-error/10 text-error">Sync Error</span>
               {{else}}
               {{statusBadge (printf "%s" .Status)}}
               {{end}}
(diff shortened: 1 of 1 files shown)


canalesb93 (commented):
This seems like we're showing too much redudnant information. Needs clean-up (or slight redesign).

canalesb93:
Good call — pushed a cleanup in 15c81b6:

- Dropped the `Sync Error` pill in the top-right. The red inline banner (color + alert icon) already signals the error state, so the pill was saying the same thing twice.
- Removed the `View Details` button inside the banner. The whole card header is already a link to the connection page, so the button was redundant with the card-level link (reauth's banner keeps its button because `Re-authenticate` is a distinct action).

Result: one banner, one link target. The earlier after-screenshots in the PR body show the old (redundant) state — treat them as stale.

PR #660, opened 2026-04-23
feat(dashboard): add compact overview strip with shortcut links

canalesb93:
## Summary

Follow-up to #657. Adds a small top-of-dashboard overview strip with **Net Worth**, **Uncategorized**, and **Pending Reviews** — the latter two link to their queues. A fourth **Connection Errors** tile appears only when there are connection errors. Uses existing `DashboardProps` fields, so no new queries.

### Desktop (1280×800)

dashboard — desktop with overview stats

### Mobile (390×844)

dashboard — mobile with overview stats

## Changes

- New `dashOverviewStats` templ component, rendered above the Agent Reports / Recent Transactions row.
- Responsive grid: `grid-cols-2` on mobile (Net Worth spans full width, the two action tiles share the second row), `lg:grid-cols-3` on desktop (even three-column row).
- Empty-state branch unchanged — when the user has no accounts, the "Connect a Bank" CTA shows instead.

## Test plan

- [x] `go build ./...` clean
- [x] `templ generate` + `templ fmt`
- [x] `make css` (new `lg:col-span-1` class needed JIT pickup)
- [x] Loaded `/` in Chrome DevTools MCP at 1280×800 and 390×844, verified tiles render and links navigate to `/transactions?filter=uncategorized` and `/reviews`.


canalesb93:
Follow-up: merged the `Total Transactions` + `Uncategorized` tiles into a single **Transactions** tile that mirrors the **Net Worth** tile's pattern — primary number for the total, smaller secondary stat for the uncategorized count. The tile title links to `/transactions`; the inline `Uncategorized N` stat links to `/transactions?filter=uncategorized`.

Grid drops from 4 action tiles to 3 (`lg:grid-cols-3`). When both `Pending Reviews` and `Connection Errors` are present, the error tile spans the full row to avoid an orphan in a 3-col grid.

### Desktop (1280×800)

dashboard — desktop with merged transactions tile

### Mobile (390×844)

dashboard — mobile with merged transactions tile

PR #659, opened 2026-04-23
feat(ui): add Alpine $store.device with isTouch/isMobile

canalesb93:
First step of the keyboard-navigation sprint: registers an Alpine `$store.device` with `isTouch` and `isMobile` so subsequent PRs (kbd-hint component, shortcut dispatcher) can branch on input capability without re-running `matchMedia` at every call site.

## Summary
- Adds a new `<script>` block in `internal/templates/layout/base.html` that, on `alpine:init`, populates `Alpine.store('device', { isTouch, isMobile })`.
- `isTouch` derives from `matchMedia('(hover: none) and (pointer: coarse)')` — true on phones, false on most laptops/desktops including those with touchscreens.
- `isMobile` derives from `matchMedia('(max-width: 1023.98px)')` — matches the `lg:` breakpoint the drawer / mobile navbar already use.
- Detection is evaluated once; no reactive listeners. Nothing in the UI currently needs mid-session transitions and keeping the store static avoids forcing consumers into reactive getters.
- No visible UI change. The store has no callers yet — M0.2 (shortcut registry) and M0.4 (kbd templ component) will introduce them.

## Sprint plan
Tracked in `~/Documents/obsidian/Breadbox/planned-features/keyboard-nav-sprint.md` → **M0.1** of the `keyboard-nav` stack.

## Test plan
- [ ] `go build ./...` passes (verified locally).
- [ ] `go vet ./...` passes (verified locally).
- [ ] In browser DevTools console on any admin page: `Alpine.store('device')` returns `{ isTouch: false, isMobile: false }` on desktop; emulating a mobile device flips both to `true`.

PR #661, opened 2026-04-23
feat(ui): add Alpine shortcuts registry + registry-driven dispatcher

canalesb93:
Adds the Alpine shortcuts registry (`$store.shortcuts`) and rewrites the global keydown dispatcher in `base.html` to route through it — no observable behavior change.

## Why

M0.2 in the [keyboard-nav sprint](https://github.com/canalesb93/breadbox/pull/659). Before we can make the `?` help modal (M0.3) and Cmd+K command palette (M0.5) reflect page-scoped shortcuts, we need a single source of truth they can both read from. This PR lands that registry + dispatcher; follow-ups swap each consumer over.

## What changed

**New head script — `Alpine.store('shortcuts')`:**
- `register(spec)` — idempotent by `id`; re-registering replaces so Alpine re-init doesn't double-bind.
- `unregister(id)`.
- `setScope(s)` / `currentScope` — pages call `setScope('transactions')` in `x-init`.
- `forScope(scope)` — returns global ∪ scope specs (feeds help modal in M0.3).
- Spec shape documented inline: `{ id, keys, description, group, scope, when?, action, visible?, allowInInput? }` where `keys` accepts `'c'`, `['c']`, `{ chord: ['g','d'] }`, or `'cmd+k'`.

**Dispatcher rewrite at the bottom of `base.html`:**
- Same input-focus / overlay / modifier guards as before.
- Chord resolution: 800ms window, cleared on any miss or success — identical to old timing.
- Single-key resolution against registered specs.
- **Touch-device guard**: non-global specs skip when `$store.device.isTouch`. Globals still fire so Bluetooth-keyboard users on iPads retain `?`, `/`, `Cmd+K`, `g+_`, `n+_`.
- Cmd+K routing stays in `commandPalette().handleGlobalKey()` (it needs access to the palette's own state to toggle open↔closed). Registered here purely so the help modal / palette can surface the binding.

**Seed:** on `alpine:init`, a deferred `seedRegistry()` registers today's exact bindings — `?`, `/`, `cmd+k`, all 13 `g+<x>` routes, all 4 `n+<x>` routes. `goRoutes` and `newRoutes` stay as plain maps so M0.5 can reuse them verbatim for palette entries.

## Behavior preservation

Scope was deliberately tight — this PR **must not change** the UX. Manually traced through each path:

- `?` → `open-shortcuts` event fires (unchanged).
- `/` → `open-cmdk` event fires (unchanged).
- `Cmd+K` / `Ctrl+K` → handled by the palette component (unchanged).
- `g` then one of `d/t/c/r/u/a/l/b/s/p/m/k/f` within 800ms → `navigateTo(path)` with the same fade-out transition as before.
- `n` then one of `c/i/k/u` → same.
- `g` or `n` followed by an unmapped key → prefix clears silently, no `preventDefault` (matches old behavior).
- `Esc` on overlays → handled inline on each overlay via `@keydown.window.escape` (untouched).

## Out of scope (deliberate)

- M0.3 — help modal reads from registry.
- M0.5 — command-palette rows render kbd hints.

Help modal and Cmd+K palette continue to render their hardcoded lists so the diff stays tight and each surface gets its own reviewable PR.

## Test plan

- [x] `go build ./...` clean.
- [x] `go vet ./...` clean.
- [ ] Verify `?`, `/`, `Cmd+K` still work on a desktop.
- [ ] Verify `g d`, `g t`, `n c` still navigate.
- [ ] Verify typing `/` inside a text input doesn't trigger the palette.
- [ ] Verify `Esc` still closes the command palette and shortcuts help.

### DIFF
--- a/internal/templates/layout/base.html
+++ b/internal/templates/layout/base.html
@@ -75,6 +75,66 @@
       });
     });
   </script>
+
+  <script>
+    // Alpine store: $store.shortcuts
+    //
+    // Central registry for all keyboard shortcuts. Pages declare their bindings
+    // via $store.shortcuts.register() on init; the global dispatcher (see the
+    // keydown listener near the bottom of this file) reads from this registry
+    // to route events. Upcoming PRs will also feed the help modal (M0.3) and
+    // command-palette kbd hints (M0.5) from the same source.
+    //
+    // Spec shape (see planned-features/keyboard-nav-sprint.md):
+    //   {
+    //     id:          string (unique),
+    //     keys:        'c' | ['c'] | { chord: ['g','d'] } | 'cmd+k',
+    //     description: short user-facing string,
+    //     group:       'Navigation' | 'Go to' | 'New' | 'Actions' | 'Selection',
+    //     scope:       'global' | route id (e.g. 'transactions'),
+    //     when:        optional () => bool,
+    //     action:      () => void | { cmdk: { label, icon, href } } palette-only,
+    //     visible:     default true (hide from help modal; action still wired),
+    //     allowInInput: optional bool (shortcut still fires while typing),
+    //   }
+    //
+    // M0.2 scope: register + dispatch only. The help modal and command palette
+    // remain hardcoded in this PR; they will be rewired to read from this
+    // registry in M0.3 and M0.5.
+    document.addEventListener('alpine:init', function() {
+      Alpine.store('shortcuts', {
+        items: [],
+        currentScope: 'global',
+
+        register(spec) {
+          if (!spec || !spec.id) return;
+          // Idempotent by id — re-registering replaces the previous spec so
+          // Alpine re-init (htmx boosts, drawer re-render) doesn't double-bind.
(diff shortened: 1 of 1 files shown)


canalesb93:
* **#661** <a href="https://app.graphite.com/github/pr/canalesb93/breadbox/661?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/canalesb93/breadbox/661?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#659** <a href="https://app.graphite.com/github/pr/canalesb93/breadbox/659?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

canalesb93:
### Merge activity

* **Apr 23, 9:03 PM UTC**: @canalesb93 merged this pull request with [Graphite](https://app.graphite.com/github/pr/canalesb93/breadbox/661).

PR #669, opened 2026-04-23
feat(reviews): keyboard shortcuts in review queue

canalesb93:
**Scope trimmed after review feedback:** reject and skip shortcuts have been removed; only `a` approve remains. Rationale in the commit and in the review-thread reply — reject had no data-model meaning (the queue is tag-backed, so there is no distinct "rejected" state), and skip was duplicative of `j` without a tracked snooze mechanism.

M3 of the keyboard navigation sprint (see `planned-features/keyboard-nav-sprint.md`). The review queue view (`/transactions?tags=needs-review`, the target of the `/reviews` admin alias) now registers a review-specific keyboard shortcut on top of the transactions list's existing j/k/Enter/c navigation.

## Summary
- New shortcut registered at `scope: 'reviews'`:
  - `a` approve — DELETE `needs-review` tag on the focused row and advance focus.
- Page wraps a templ component (`txReviewQueueShortcuts`) that only renders when `p.IsReviewQueue()` is true; `x-init` / `x-destroy` flip the shortcut-registry scope between `reviews` and `global`.
- The action calls the same `DELETE /-/transactions/:id/tags/needs-review` endpoint the tag chip's remove button uses — CSRF, annotations, and row-removal animation stay in one code path. No new API routes, no duplicated POST logic.
- j/k/Enter/c continue to fire via the existing inline keydown handler in `transactions_scripts.js.html` until M1 lands and ports those to the registry; M3 only layers the review-specific approve action.

## Scope notes
- The review system is tag-backed since the cutover migration (`20260415083233_cutover_review_queue.sql`) dropped the `review_queue` table — approve resolves the tag.
- Registered shortcuts surface in the `?` help modal and command palette once M0.3 / M0.5 land; on main today the hardcoded help body doesn't list them yet but the key fires correctly.

## Test plan
- [ ] Navigate to `/reviews` (redirects to `/transactions?tags=needs-review`); confirm j/k focus a row, then:
  - [ ] `a` removes the needs-review tag and advances focus, toast reads "Review approved".
- [ ] On a regular `/transactions` view (no tag filter), confirm `a` does nothing (scope stays `global`).
- [ ] `go build ./...` + `go vet ./...` clean (verified locally).
- [ ] Touch device: page-scoped shortcuts are suppressed by the dispatcher's `isTouch` guard — no new handler bypasses it.

### DIFF
--- a/internal/templates/components/pages/assets/reviews_scripts.js.html
+++ b/internal/templates/components/pages/assets/reviews_scripts.js.html
@@ -2,17 +2,18 @@
  * M3 — Reviews queue keyboard shortcuts.
  *
  * Rendered only when the transactions page is filtered to tags=needs-review
- * (see txReviewQueueShortcuts in transactions.templ). Registers the three
- * review-specific actions — approve (a), reject (r), skip (s) — at
- * `scope: 'reviews'` on top of the transactions list's existing j/k/Enter/c
- * handlers (still driven by the inline keydown in transactions_scripts.js.html
- * until M1 lands and ports them to the registry).
+ * (see txReviewQueueShortcuts in transactions.templ). Registers the single
+ * review-specific action — approve (a) — at `scope: 'reviews'` on top of the
+ * transactions list's existing j/k/Enter/c handlers (still driven by the
+ * inline keydown in transactions_scripts.js.html until M1 lands and ports
+ * them to the registry).
  *
- * Approve/reject both resolve the needs-review tag on the focused row: the
- * review system is tag-backed since the review_queue cutover (#... migration
- * 20260415083233_cutover_review_queue.sql) so there is no distinct "rejected"
- * state in the data model — only a different annotation / toast. Skip leaves
- * the tag intact and just advances focus so the reviewer can come back later.
+ * Reject and skip were considered but dropped: the review model is tag-backed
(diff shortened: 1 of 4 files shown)


canalesb93:
I think the actions are not quite what we want. Reject and skip don’t really make sense unless we do some action with them. 

Reject makes least sense. What does it mean to reject a transaction? Overall it’s not a valid action IMO. 

Skip could mean skip to later. But not sure how that would be tracked in this review view. 

We could leave the approve action but the others idk tbd. In general the transaction list should not have custom logic and should fit all use cases. “Review queue” might be the exception but still needs to meet high standard to excuse the change.

canalesb93:
Good call — trimmed the scope in 33a2ad5:

- **Removed `r` reject**: you're right, reject has no data-model meaning. The queue is tag-backed since the cutover migration, so there's no distinct "rejected" state — it was just approve with different toast copy. Gone.
- **Removed `s` skip**: agreed, without a tracked "snooze until later" mechanism it's literally just `j` (advance focus). Gone.
- **Kept `a` approve**: the one action that actually resolves the review (DELETE `needs-review` tag + advance focus + toast). This doesn't add custom list logic — it just binds a key to an action the tag chip already does via mouse.

PR body updated to match. No new API routes, no duplicated handlers — the transactions list stays generic.

Issue #687, opened 2026-04-24
cli: add `breadbox doctor` pre-flight + readiness command


canalesb93:
## Summary

Add a `breadbox doctor` subcommand that surfaces configuration and connectivity issues **without booting the server**. Today, startup validation at `cmd/breadbox/main.go:185-188` only fails fast when a provider is configured but `ENCRYPTION_KEY` is missing. Everything else — bad DATABASE_URL, unreadable Teller cert, misconfigured Plaid env, missing admin account, unreachable public URL — is discovered at runtime, often long after `serve` started.

## Proposed checks

- `DATABASE_URL` set, Postgres reachable, migrations applied (compare `goose_db_version` vs embedded migrations)
- `ENCRYPTION_KEY` set, 32 bytes when hex-decoded, matches what's encrypting existing connections (compare key fingerprint vs a stored hash — pairs with the encryption-key-UX issue)
- If any provider configured: credentials decrypt cleanly; provider env sane (Plaid env, Teller cert/key paths exist and are readable)
- Admin account exists (`auth_accounts` count > 0)
- If `DOMAIN` / `PUBLIC_URL` set: DNS resolves, `/health/ready` reachable from outside — skip with a `--skip-external` flag for air-gapped installs
- Scheduler config valid (cron syntax parses)

## Shape

- Same binary: `breadbox doctor`
- stdout pretty table (pass / fail / skip per check, with remediation hints)
- Non-zero exit on any check failure so install scripts can gate on it
- Structured output mode (`--json`) for CI

## Why

Paired with the install-UX issue — `doctor` is the readiness-check surface the new one-liner installer would call after launching, and the command users are told to run when something feels off.

canalesb93:
Picking this up — a Claude Code subagent is starting on it now.

PR #697, opened 2026-04-24
feat(crypto): print-once key + wizard confirmation + fingerprint surface (#688)

### LINKED ISSUE #688: crypto: improve encryption-key UX (print-once, wizard confirmation, key fingerprint)
## Summary

The encryption-key UX has a silent-failure cliff: if the user loses `.env`, every bank connection becomes un-decryptable ciphertext and recovery requires re-linking every institution. We should print the key once with a clear warning, confirm it in the setup wizard, and surface a key fingerprint across the admin UI so migrations have a visible sanity anchor.

Full proposal: `docs/proposals/encryption-key-ux.md` (currently untracked locally — I'll commit separately).

## Current flow

- `install.sh` generates `ENCRYPTION_KEY`, writes it to `.env`, and never surfaces it again.
- If the user loses `.env` — `rm`, host migration without copying the file, Docker volume rebuild — all `connections.encrypted_credentials` become AES-256-GCM ciphertext the server can't decrypt.
- Recovery path = unlink and re-link every institution (Plaid + Teller), losing historical sync state.

### Silent failure modes

- **F1**: `.env` deleted. Startup emits a `Warn` line (easy to miss), then GCM auth-failed errors start showing up at the first sync.
- **F3**: host migration without copying `.env`. Same shape — no explicit "your key is wrong" surface, just decryption errors on first real work.

### Phase 1 — immediate

- `install.sh` prints the generated `ENCRYPTION_KEY` once with a bright "SAVE THIS NOW" warning, boxed off from the rest of the output.
- Extend `internal/admin/setup.go` with a pre-step: show the key + its fingerprint, require a "I've saved my encryption key" checkbox before the wizard continues.
- Add a **key fingerprint** (first 8 chars of `sha256(key)`) to the Settings page and the Backups page, so host migrations and `.env` restores have a visible anchor to sanity-check against.
- Pairs with `breadbox doctor`: the doctor compares the current key's fingerprint against a fingerprint stored alongside `connections.encrypted_credentials` (or in `app_config`) and fails loudly on mismatch.

### Phase 2

- Optional 24-word BIP39 mnemonic encoding of the same 32-byte key, printed alongside the hex form during install and downloadable from Settings. Humans who prefer paper backups get one; the hex form still works for scripts.

### Phase 3 — separable

- `breadbox rotate-key` subcommand. Reads both old and new keys, re-wraps every row in `connections.encrypted_credentials`, updates the stored fingerprint, writes out a new `.env`. Runs in a transaction so a crash mid-rotate doesn't brick the install.

## Rejected alternative

- **(d) unseal model** (Vault-style — server boots sealed, admin unseals with key on every start). Too aggressive for Breadbox's unattended-sync workflow; cron-driven syncs would stop on every reboot until a human unsealed the box.

## Scope guard

Phase 1 is the minimum bar and unblocks the install-UX issue. Phase 2 and Phase 3 are independent follow-ups and can ship separately.

canalesb93:
Closes #688.

Phase 1 of the encryption-key UX work. Operators now have a visible anchor on `ENCRYPTION_KEY` at every point where the key matters: install, first-run wizard, Settings, and Backups.

## What's in the box

- **`deploy/install.sh`** — prints the generated key once in a boxed, loud "SAVE THIS NOW" panel with its 8-char fingerprint. A rerun against an existing install surfaces the *existing* key's fingerprint instead of regenerating. The final success block also re-prints the fingerprint alongside the setup-wizard URL.
- **`internal/crypto/fingerprint.go`** — the only place fingerprint derivation lives. `Fingerprint(key) = first 8 hex of sha256(key)`. install.sh, server startup, and every UI site call this so the value an admin writes on paper at install matches what they see later in the dashboard.
- **Setup wizard pre-step** (`/setup/key`) — shows the fingerprint and requires an "I've saved my encryption key" checkbox before `/setup` renders. Gated on GET only so programmatic `POST /setup` and `POST /-/setup` flows (tests, headless installs) keep working unchanged.
- **Settings → Security card** — surfaces the fingerprint under the existing "Encryption Key / Active" status row, with explanatory copy that points at host migrations and `.env` restores as the thing you're guarding against.
- **Backups page** — adds the fingerprint to the "Back up your ENCRYPTION_KEY separately" warning so operators can confirm the target host's key matches before restoring a backup.
- **TOFU storage** — startup writes the observed fingerprint to `app_config.encryption_key_fingerprint` on first boot. A mismatch on subsequent boots logs a warning; #687's `breadbox doctor` will upgrade that into a hard stop once we have a supported rotation path.

## Storage decision

I put the fingerprint in `app_config` rather than a new table or a column on `connections`. Rationale:

- It's a single server-wide KV value — `app_config` is already the write surface for every other server-managed setting (`backup_schedule`, `sync_interval_minutes`, onboarding flags).
- A dedicated table or migrated column would require schema changes for every future version; `app_config` absorbs the churn.
- Pairs cleanly with `appconfig.String(..., "encryption_key_fingerprint", "")` which is the idiomatic read path elsewhere.

The migration is additive: `INSERT ... ON CONFLICT (key) DO NOTHING`, so it's safe to replay on the shared dev DB without clobbering a running server's fingerprint.

## Out of scope (deferred follow-ups)

- **Phase 2** — BIP39 mnemonic encoding of the key.
- **Phase 3** — `breadbox rotate-key` subcommand.
- **`breadbox doctor`** (#687) — will diff observed vs stored fingerprint and fail loudly. This PR exposes and stores the values; the diff/fail comes in #687.

## Testing

- `go test ./...` — all pass, including new `TestFingerprint_*` cases.
- `make test-integration` — all pass against `breadbox_test` on the shared dev DB (no destructive migration operations).
- Verified the TOFU write: started a fresh server against `breadbox_test`, confirmed `SELECT value FROM app_config WHERE key='encryption_key_fingerprint'` returns `a39cc8bc` matching `printf <KEY_HEX> | xxd -r -p | openssl dgst -sha256 | cut -c1-8`.
- End-to-end wizard flow: `/setup` → redirected to `/setup/key` → checked the box → continued to `/setup` (admin creation). Session flag correctly gates the pre-step.

## Screenshots (Chrome DevTools MCP)

**Setup wizard — encryption-key confirmation step** (`/setup/key`)

setup/key confirmation — fingerprint + acknowledgement checkbox

**Settings → Security card** (`/settings#security`)

Settings Security card with key fingerprint

**Backups page — warning banner with fingerprint** (`/backups`)

Backups page with key fingerprint in warning banner

## Test plan

- [ ] `go test ./...`
- [ ] `make test-integration`
- [ ] Fresh install: `./deploy/install.sh` prints the boxed `SAVE THIS NOW` key panel and a matching fingerprint
- [ ] Rerun on an existing install surfaces the existing fingerprint
- [ ] `/setup/key` renders the correct fingerprint on a fresh DB and gates `/setup`
- [ ] `/settings` Security card shows the fingerprint and copy
- [ ] `/backups` warning banner shows the fingerprint and copy


canalesb93:
Rejected with feedback added in the tracker issue.

PR #699, opened 2026-04-24
refactor(kbd): show page shortcuts above globals in ? modal

canalesb93:
## Summary

Flips the `?` keyboard shortcuts help modal so the page-scoped section renders **above** the global section. Previously globals were listed first and page actions lived at the bottom, where users on a page with its own bindings (Transactions, Rules, Transaction Detail) had to scroll past a long list of globals they already know to discover the page-specific shortcuts that are actually new.

## Changes

- `internal/templates/layout/base.html`:
  - Moved the page-section `<template x-if="sections.page.length > 0">` block above the global loop in `.bb-shortcuts-body`.
  - The page section is now hidden entirely when `sections.page` is empty — e.g. current scope is `'global'` or no page-scoped shortcuts were registered. Previously an empty "This page" heading could still render.
  - Page-section heading now reflects the current scope instead of the generic "This page". `scopeLabel(scope)` derives a display name by titlecasing the kebab-case scope id: `'transactions'` → `'Transactions'`, `'transaction-detail'` → `'Transaction detail'`. Falls back to `'This page'` for `'global'` or empty. No explicit scope→name map — just a heuristic on the id, so pages need no extra metadata.

No changes to the shortcut registry, dispatcher, or any page templates. `templ generate`, `go build ./...`, `go vet ./...` all clean.

## Test plan

- [ ] `?` on `/` (global scope): only the Global section renders; no "This page" heading.
- [ ] `?` on `/transactions`: page section renders first, labeled "Transactions"; Global section below.
- [ ] `?` on a transaction detail page: page section labeled "Transaction detail".
- [ ] Escape still closes the modal.

### DIFF
--- a/internal/templates/layout/base.html
+++ b/internal/templates/layout/base.html
@@ -283,25 +283,16 @@
         </button>
       </div>
       <div class="bb-shortcuts-body">
-        <!-- Global section: shortcuts available on every page. -->
-        <template x-for="group in sections.global" :key="'global-' + group.label">
-          <div class="bb-shortcuts-group">
-            <div class="bb-shortcuts-group-label" x-text="group.label"></div>
-            <template x-for="item in group.items" :key="item.id">
-              <div class="bb-shortcuts-row">
-                <span class="bb-shortcuts-label" x-text="item.description"></span>
-                <span class="bb-shortcuts-keys" x-html="renderTokens(item.tokens)"></span>
-              </div>
-            </template>
-          </div>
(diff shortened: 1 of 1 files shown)


canalesb93:
### Merge activity

* **Apr 24, 8:02 AM UTC**: @canalesb93 merged this pull request with [Graphite](https://app.graphite.com/github/pr/canalesb93/breadbox/699).

PR #740, opened 2026-04-24
fix(rules): show retroactive-apply error inside modal, use toast for success (#701)

### LINKED ISSUE #701: ux(rules): retroactive-apply error alert renders behind modal backdrop — users can't see why apply failed
## Route + viewports
`/rules/:id` → "Categorize as …" modal → click "Apply to N" on a rule whose backend call returns an error. Bug exists at every viewport (375–1440 px); visible only via code read + DOM inspection because the error element is visually occluded.

## Current
`rule_detail.html` renders the retroactive-apply success/error alerts **outside** the modal and **before** the `x-teleport="body"` modal template:

```html
<!-- lines 333–338 -->
<div x-show="applyResult" ...class="alert alert-success mt-3">…</div>
<div x-show="applyError"  ...class="alert alert-error   mt-3">…</div>

<!-- lines 340–379: the modal, teleported to <body> with z-50 backdrop -->
<template x-teleport="body"> <div x-show="showApplyModal" ...> … </div> </template>
```

When `applyRetroactively()` catches an error (line 548), it sets `applyError` but does **not** close the modal. Because the alert lives inside the preview card (underneath the modal's `bg-black/40` z-50 overlay), the user sees the modal still open with no feedback; the error is sitting invisibly behind the backdrop. When they dismiss the modal, the error is already 0% visible in the flow and may be scrolled off-screen.

Success path is also fragile: on success the modal closes, an alert flashes for ~1.5 s, then `location.reload()` wipes the page — if the user looks away or the page finishes reloading they never see "Applied to N transactions."

## Expected
- Errors surface **inside** the modal (or a toast that floats above the modal), so the user can read them while the modal is still open and decide what to do.
- Successes render as a persistent toast / confirmation banner that survives the page reload (or the reload is skipped in favor of an in-place update).

## Repro
1. Log in as admin, open a rule detail page with ≥ 1 matching transaction.
2. Click "Categorize as …" to open the modal.
3. Check the confirm checkbox and click "Apply to N."
4. If the server returns 4xx/5xx (e.g. transient DB error, stale rule, CSRF mismatch), the modal stays open with no visible feedback. The `alert alert-error` is rendered behind the `bg-black/40` modal backdrop on the preview card.

Demonstrated in this thread by intercepting the POST and forcing a 500: the modal remains identical to its initial state with no inline error.

## Proposed fix
**File:** `internal/templates/pages/rule_detail.html:333-378`

**Change:** Move the error alert *into* the modal, and move the success alert into a viewport-level toast so it survives the reload window.

```html
<!-- Inside the modal, just above the action row (around line 368) -->
<div x-show="applyError" x-cloak role="alert"
     class="alert alert-error text-xs mt-3 rounded-xl">
  <i data-lucide="alert-circle" class="w-4 h-4"></i>
  <span x-text="applyError"></span>
</div>

<!-- Remove the old in-card error block at line 336-338 -->
<!-- Move success into the existing .toast container or add a new one, e.g. -->
<div x-show="applyResult" x-cloak
     class="toast toast-top toast-center z-[60]">
  <div class="alert alert-success text-sm rounded-xl shadow-md">
    <i data-lucide="check" class="w-4 h-4"></i>
    <span x-text="applyResult"></span>
  </div>
</div>
```

Also raise modal `z-50` → `z-40` for the backdrop and keep the toast at `z-50` so success toasts sit above any residual overlay.

**Rationale:** The current layout assumes the alert will always be visible because the modal auto-closes on success. But (a) the modal does **not** auto-close on error, and (b) the 1.5 s visibility on success is fragile. Placing the error inside the dialog and promoting success to a toast makes both paths discoverable.

## Acceptance criteria
- [ ] When `/-/rules/{id}/apply` returns a non-2xx response, an inline error alert appears **inside** the modal above the action buttons, at all 5 viewports.
- [ ] The modal stays open on error and the user can retry or cancel.
- [ ] On success, a toast persists until the page reload completes and is clearly visible at M/T/D widths.
- [ ] No regression: repeated open/close cycles clear stale `applyError`/`applyResult` (already handled in `openApplyModal`).

canalesb93:
## Summary

Fix `/rules/{id}` retroactive-apply error alert rendering behind the
modal backdrop.

- Move the error alert **inside** the modal (above Cancel/Apply) so
  users see it while the dialog is still open and can retry or cancel.
- Promote the success message to the global `bb-toast` stream so it
  survives the reload window (the old inline success alert flashed for
  ~1.5s before `location.reload()` wiped it).
- Drop the unused `applyResult` Alpine state now that success routes
  through the toast.

Fixes #701.

## UI evidence

Simulated via request interception returning `500 { error: "…" }` on
`POST /-/rules/{id}/apply`.

<table>
  <tr><th>Viewport</th><th>Before</th><th>After</th></tr>
  <tr>
    <td>Desktop (1440)</td>
    <td>before desktop</td>
    <td>after desktop</td>
  </tr>
  <tr>
    <td>Tablet (820)</td>
    <td>before tablet</td>
    <td>after tablet</td>
  </tr>
  <tr>
    <td>Mobile (500)</td>
    <td>before mobile</td>
    <td>after mobile</td>
  </tr>
</table>

On the "before" shots, the error alert is the red banner below the
preview card — visually hidden behind the z-50 modal backdrop.
On the "after" shots, it's inline inside the dialog above the action
buttons, so the user can read it, retry, or cancel without dismissing
the modal first.

## Testing

- [x] `go build ./...`
- [x] `go test ./...` passes
- [x] Manual puppeteer flow on 500/820/1440: confirm 500-response path
      surfaces inline error; no layout regression in the initial and
      modal-open states.

## Related issues

Closes #701

### DIFF
--- a/internal/templates/pages/rule_detail.html
+++ b/internal/templates/pages/rule_detail.html
@@ -359,12 +359,6 @@
     </button>
   </div>
 
-  <div x-show="applyResult" x-cloak role="alert" class="alert alert-success text-sm mt-3 rounded-xl">
-    <span x-text="applyResult"></span>
-  </div>
-  <div x-show="applyError" x-cloak role="alert" class="alert alert-error text-sm mt-3 rounded-xl">
-    <span x-text="applyError"></span>
-  </div>
   {{end}}
   {{else}}
   <!-- 0-match empty state: matches the tone of other empty cards (e.g. the
@@ -415,6 +409,15 @@
             </label>
           </div>
         </div>
+      </div>
+      <!-- Inline error alert lives *inside* the modal so it's visible while
+           the dialog is still open. The old in-card alert (above the
+           category button) was rendered behind the z-50 backdrop and
+           therefore invisible whenever apply failed — see #701. -->
+      <div x-show="applyError" x-cloak role="alert"
+           class="alert alert-error text-xs mt-4 rounded-xl">
+        <i data-lucide="alert-circle" class="w-4 h-4 shrink-0"></i>
+        <span x-text="applyError"></span>
       </div>
       <div class="flex items-center justify-end gap-2 mt-5">
         <button type="button" class="btn btn-sm btn-ghost rounded-xl" @click="closeApplyModal()">Cancel</button>
@@ -527,7 +530,6 @@ function quickSetCategory(txId, categoryId) {
 function ruleDetail() {
   return {
     applying: false,
-    applyResult: '',
     applyError: '',
     showApplyModal: false,
     confirmApply: false,
@@ -584,7 +586,6 @@ function ruleDetail() {
     openApplyModal() {
       this.confirmApply = false;
       this.applyError = '';
-      this.applyResult = '';
       this.showApplyModal = true;
     },
     closeApplyModal() {
@@ -594,7 +595,6 @@ function ruleDetail() {
     async applyRetroactively() {
       if (!this.confirmApply) return;
       this.applying = true;
-      this.applyResult = '';
       this.applyError = '';
       try {
         const resp = await fetch('/-/rules/{{.Rule.ID}}/apply', { method: 'POST' });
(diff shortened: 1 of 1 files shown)


canalesb93:
* **#740** <a href="https://app.graphite.com/github/pr/canalesb93/breadbox/740?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/canalesb93/breadbox/740?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

Issue #783, opened 2026-04-25
fix(access): full OAuth client / API key IDs are unrecoverable post-creation — only 7-char prefix shown, no copy, no detail page


canalesb93:
## Surface
Route: `/access`  |  Viewports affected: 375 / 500 / 820 / 1440

## Current behavior

Once an OAuth client or API key is created, the only on-screen identifier is a 7-character prefix followed by an ellipsis (`bb_oc_1PkkLZA1...`, `bb_pat_xxxxxxx...`). There is:

- No copy-to-clipboard button on the row.
- No detail page for an existing client/key (only `/oauth-clients/{id}/created` and `/api-keys/{id}/created`, which are one-shot post-create screens).
- No way to expand the truncated `<code>` (no hover-to-reveal, no click-to-copy).

If the user dismisses the "created" screen without copying the full client ID, they have no way to recover it from the admin UI — they have to `psql` into the dev DB or recreate the client. With multiple similar names (e.g. several `Screenshot after desktop ...` test clients), the truncated prefix isn't enough to disambiguate either.

![current](https://i.img402.dev/61jgqg35b5.png)

## Expected behavior

- Add a copy-icon button next to the `<code>` block that copies the full client ID / key prefix.
- (Optional, follow-up) provide a read-only detail page at `/oauth-clients/{id}` showing the full client ID, scope, created/last-used timestamps, and a Revoke action.

## Repro steps

1. Sign in as an admin and visit `/access`.
2. Note that every OAuth client row shows only `bb_oc_xxxxxxx...` with no way to read the full ID.
3. Try to copy the full ID — there is no UI affordance.

## Proposed fix

- File: `internal/templates/pages/access.html` (lines 46 and 152, the two `<code>...{{.ClientIDPrefix}}...</code>` / `{{.KeyPrefix}}...` blocks).
- Change: wrap the `<code>` in an Alpine `x-data="{ copied: false }"` cluster, store the full ID via `data-id="{{.ID}}"` (or expose `.ClientID` from the handler), and add a `<button class="btn btn-ghost btn-xs btn-circle" aria-label="Copy client ID" @click="navigator.clipboard.writeText($el.previousElementSibling.dataset.id); copied = true; setTimeout(() => copied = false, 1500)">` with a swap from `copy` → `check` icon. Mirror the `{{.KeyPrefix}}` row.
- Rationale: full identifiers are needed to register an external connector; the current UI hides them irretrievably after the create flow.

## Acceptance criteria

[...]

PR #813, opened 2026-04-25
fix(access): move per-row Revoke into kebab dropdown

### LINKED ISSUE #784: ux(access): inline Revoke text-button per row competes with primary content — move to kebab dropdown (#728 precedent)
## Surface
Route: `/access`  |  Viewports affected: 375 / 500 / 820 / 1440

## Current behavior

Every active OAuth client and API key row exposes a destructive **Revoke** text-button inline at the right edge of the row. With 8 active clients, that's 8 destructive triggers stacked one above the other, mixed in with primary content (name, scope badge, ID prefix, created date). The Alpine swap into `Confirm` / `Cancel` is fine, but the *trigger* itself sits in the primary attention path.

This matches the kebab/overflow-menu candidate criteria called out in the triage brief and the precedent set in [PR #728](https://github.com/canalesb93/breadbox/pull/728):
- Inline destructive action right next to primary content.
- Same affordance repeated on every row in a list.
- At 375 px the **Revoke** text-button consumes the right column and forces the client name to truncate hard (e.g. `Screenshot after desktop 17...`).
- Hover-only confirmation pattern (`x-show="!confirming"`) means there's no extra step protecting accidental clicks on touch devices — first tap shows the inline `Confirm/Cancel`, but the row already shifted.

current 1440
current 375

## Expected behavior

Replace the inline Revoke text-button with a DaisyUI `dropdown dropdown-end` + `btn-ghost btn-circle` kebab trigger (3-dots), per the same pattern PR #728 established for rules. Menu contents:

- **Copy client ID** (resolves the issue where the full ID is not readable, see also issue #_).
- **Revoke** (opens a small confirmation modal or keeps the inline `Confirm/Cancel` swap, but only after the menu item is chosen).

This keeps the row clean, makes mobile breathe, and treats Revoke with appropriate friction.

## Repro steps

1. Sign in as an admin and visit `/access` with 5+ OAuth clients.
2. Resize to 375 px — note that the Revoke button is the same size as at 1440 and pushes the client name into a hard truncate.
3. Note that on every row, a destructive action sits one click away with no visual separation from the rest of the row content.

## Proposed fix

- File: `internal/templates/pages/access.html`
- Change: in both the `ActiveClients` block (≈line 50–61) and the `ActiveKeys` block (≈line 157–168), replace the `<form>...</form>` Revoke cluster with a DaisyUI dropdown:

  ```html
  {{if $.IsAdmin}}
  <div class="dropdown dropdown-end shrink-0">
    <button tabindex="0" class="btn btn-ghost btn-circle btn-xs" aria-label="More actions" title="More">
      <i data-lucide="more-horizontal" class="w-4 h-4"></i>
    </button>
    <ul tabindex="0" class="dropdown-content menu menu-sm bg-base-100 rounded-box shadow-lg border border-base-300 z-10 w-44">
      <li><button type="button" @click="navigator.clipboard.writeText('{{.ID}}')"><i data-lucide="copy" class="w-3.5 h-3.5"></i>Copy client ID</button></li>
      <li><button type="button" @click="confirming = true" class="text-error"><i data-lucide="trash-2" class="w-3.5 h-3.5"></i>Revoke…</button></li>
    </ul>
  </div>
  <form method="POST" action="/oauth-clients/{{.ID}}/revoke" x-show="confirming" x-cloak class="...">…inline confirm…</form>
  {{end}}
  ```

  Mirror the same change for `ActiveKeys`. Reference PR #728 for the exact kebab + dropdown styling already used elsewhere.
- Rationale: protects against accidental destructive clicks, frees mobile horizontal space, matches the design-system precedent.

## Acceptance criteria

- [ ] Before/after screenshots at 375, 500, 820, 1440 embedded in PR body
- [ ] Screenshots uploaded to img402.dev
- [ ] `go build ./...` passes
- [ ] No regression at unaffected viewports
- [ ] Kebab menu is keyboard-operable (Tab to focus, Enter/Space to open, Escape to close)
- [ ] At 375 px, OAuth client names get noticeably more horizontal space than today

canalesb93:
Fixes #784

Replace the inline **Revoke** text-button on every OAuth client / API key row with a single kebab dropdown (`btn-ghost btn-circle` + `ellipsis-vertical`), matching the precedent set in #728. The destructive action moves out of the primary attention path and onto a deliberate menu choice, which gives client names noticeably more horizontal space at narrow widths and removes a row of repeated red triggers from the page.

The kebab menu exposes:

- **Copy client ID** — pulls the full `client_id` from the existing `OAuthClientResponse.ClientID` field. (For API keys, only the prefix is server-side recoverable, so the menu offers **Copy key prefix**; the full key is only shown once at creation by design.)
- **Revoke…** — opens the existing inline `Confirm / Cancel` swap (Alpine `confirming` flag) so the actual revoke still requires a second deliberate click.

The dropdown is keyboard operable (Tab focuses the trigger, DaisyUI opens it on focus, Enter activates a menu item, Escape / blur closes). `aria-label` includes the client / key name so screen-reader users get row context.

## Before
<table>
<tr><td><strong>375</strong></td><td><strong>500</strong></td><td><strong>820</strong></td><td><strong>1440</strong></td></tr>
<tr>
<td>before 375</td>
<td>before 500</td>
<td>before 820</td>
<td>before 1440</td>
</tr>
</table>

## After
<table>
<tr><td><strong>375</strong></td><td><strong>500</strong></td><td><strong>820</strong></td><td><strong>1440</strong></td></tr>
<tr>
<td>after 375</td>
<td>after 500</td>
<td>after 820</td>
<td>after 1440</td>
</tr>
</table>

### Dropdown opened (keyboard focus on first row trigger, 1440 px)

dropdown open showing Copy client ID and Revoke

## Notes

- The proposed fix in the issue suggested mirroring `Copy client ID` for API keys; substituted **Copy key prefix** since the full plaintext key isn't recoverable post-creation (hashed at rest). Same affordance shape, honest content. A future PR can revisit this when #783 (full-ID recovery) is addressed.
- API Keys section in the dev DB happens to be empty in the captured screenshots, so only the OAuth Clients list visibly demonstrates the change. The same template block was edited for both, and the pattern is identical.
- Empty-state cards and the revoked-clients collapsed list are unaffected.
- `go build ./...` passes.

### DIFF
--- a/internal/templates/components/pages/access.templ
+++ b/internal/templates/components/pages/access.templ
@@ -105,16 +105,27 @@ templ accessActiveClientRow(c AccessClientRow, isAdmin bool, csrfToken string) {
 			</div>
 		</div>
 		if isAdmin {
-			<form method="POST" action={ templ.SafeURL(accessOAuthClientRevokeURL(c.ID)) } class="shrink-0">
-				<input type="hidden" name="_csrf" value={ csrfToken }/>
-				<button type="button" class="btn btn-ghost btn-xs rounded-lg text-base-content/40 hover:text-error" x-show="!confirming" @click="confirming = true">
-					Revoke
-				</button>
-				<span x-show="confirming" x-cloak class="flex items-center gap-1.5 justify-end">
-					<button type="submit" class="btn btn-error btn-xs btn-soft rounded-lg">Confirm</button>
-					<button type="button" class="btn btn-ghost btn-xs rounded-lg" @click="confirming = false">Cancel</button>
-				</span>
-			</form>
+			<div class="shrink-0 flex items-center">
+				<div class="dropdown dropdown-end" x-show="!confirming">
+					<div tabindex="0" role="button" class={ "btn btn-ghost btn-xs btn-square rounded-lg text-base-content/40" } aria-label={ "Actions for OAuth client " + c.Name } title="More actions">
+						<i data-lucide="ellipsis-vertical" class="w-4 h-4"></i>
+					</div>
+					<ul tabindex="0" class="dropdown-content menu menu-sm bg-base-100 rounded-xl shadow-lg border border-base-300 z-50 w-44 p-1">
+						<li>
+							<button type="button" class="text-error" @click="confirming = true; document.activeElement.blur()">
+								<i data-lucide="trash-2" class="w-3.5 h-3.5"></i> Revoke&hellip;
+							</button>
+						</li>
+					</ul>
+				</div>
+				<form method="POST" action={ templ.SafeURL(accessOAuthClientRevokeURL(c.ID)) } x-show="confirming" x-cloak>
+					<input type="hidden" name="_csrf" value={ csrfToken }/>
+					<span class="flex items-center gap-1.5 justify-end">
+						<button type="submit" class="btn btn-error btn-xs btn-soft rounded-lg">Confirm</button>
+						<button type="button" class="btn btn-ghost btn-xs rounded-lg" @click="confirming = false">Cancel</button>
+					</span>
+				</form>
+			</div>
 		}
 	</div>
 }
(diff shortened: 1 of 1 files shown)


canalesb93:
* **#813** <a href="https://app.graphite.com/github/pr/canalesb93/breadbox/813?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/canalesb93/breadbox/813?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #823, opened 2026-04-25
fix(admin): align activity timeline avatars with user identity

### LINKED ISSUE #788: Activity timeline avatars don't resolve for admin actors (actor_id stores auth_accounts.id, not users.id)
Follow-up to #781 (PR notes).

## Bug

The activity timeline never resolves an uploaded avatar for actions taken by an admin. After admin avatar upload, the timeline keeps rendering the generated pattern even though `/avatars/{users.id}` would serve the uploaded image.

## Root cause

`annotations.actor_id` is populated from `service.Actor.ID`, which `ActorFromSession` (`internal/admin/auth.go:129`) sets to the **`auth_accounts.id`** for admin sessions:

```go
return service.Actor{Type: "user", ID: id, Name: name}
// where id = sm.GetString(r.Context(), sessionKeyAccountID)
```

But the avatar handler at `GET /avatars/{id}` (`internal/admin/avatars.go:28`) keys lookups on **`users.id`** via `GetUserAvatar`. When the URL `id` is an `auth_accounts.id`, `GetUserAvatar` misses and the handler falls through to `serveGeneratedAvatar`, even when the underlying linked user has uploaded a real avatar.

#781 mitigated this for the cache-busting layer by JOINing `auth_accounts → users` in `ListAnnotationsWithActorByTransaction` to pick up `users.updated_at` regardless of which UUID flavor is in `actor_id`. The actor-id mismatch itself is unchanged, so timeline `` URLs still point at `/avatars/{auth_accounts.id}` and 404 into the generated pattern.

## Fix options

1. **Resolve at write time** — when `ActorFromSession` builds the `service.Actor`, prefer the linked `users.id` (already stashed on login at `internal/admin/auth.go:120` via `sessionKeyUserID`) and only fall back to `auth_accounts.id` for unlinked initial admins. Cleanest; future annotations get the right id. Existing rows stay broken unless backfilled.

2. **Resolve at read time** — `buildActivityTimeline` (and any other consumer that turns `actor_id` into an avatar URL) should look up `auth_accounts.user_id` and emit `/avatars/{users.id}` when present. The JOIN added in #781 already brings the right user into scope; this would just plumb it.

3. **Resolve in the handler** — `AvatarHandler` could fall through `auth_accounts.id → users.id` before generating a pattern. Lowest blast radius but adds a query per avatar request.

Option 1 + a one-shot backfill of `annotations.actor_id` for `actor_type='user'` rows where `actor_id` matches an `auth_accounts.id` is probably the right shape.

## Acceptance

- Admin uploads an avatar at `/account`.
- Open a transaction whose timeline contains an admin-authored entry (comment, tag change, category set).
- The avatar in that timeline row renders the uploaded image, not the generated pattern.

canalesb93:
Fixes [#788](https://github.com/canalesb93/breadbox/issues/788) — activity-timeline rows showed the auto-generated pattern avatar instead of the admin's uploaded gradient.

## Root cause

`actor_id` on annotations was being written as `auth_accounts.id`, but the avatar handler at `/avatars/{id}` keys lookups on `users.id`. The IDs differ on linked admin accounts, so the timeline silently fell through to `serveGeneratedAvatar` while the sidebar (which already uses `users.id`) rendered the real picture. Same admin, two avatars.

While fixing the actor side I also hit a pre-existing bug in `ListAnnotationsWithActorByTransaction` (added in #781): the JOIN cast `actor_id::uuid` blew up on `rule_applied` annotations that store the rule's 8-char `short_id` in `actor_id` (`SQLSTATE 22P02`). The activity timeline silently rendered "0 events" for any transaction whose history included a rule application.

## Fix

1. `ActorFromSession` prefers `sessionKeyUserID` over `sessionKeyAccountID`, falling back only for unlinked initial admin accounts. New annotations attach to the user identity.
2. `ListAnnotationsWithActorByTransaction` compares in text space (`aa.id::text = a.actor_id`) rather than casting `actor_id::uuid`. Postgres does not guarantee short-circuit evaluation in JOIN conditions, so the previous regex guard alone wasn't enough.
3. One-shot backfill migration rewrites existing `annotations.actor_id` rows from `auth_accounts.id` to the linked `users.id`.

## Validation

| | Before | After |
|---|---|---|
| Desktop (1440×900) | before — desktop | after — desktop |
| Tablet (820×1180) | before — tablet | after — tablet |
| Mobile (500×844) | before — mobile | after — mobile |

Notice the sidebar avatar (gradient) matches the timeline avatar in the After column — they were diverging before.

## Test plan

- [x] `go build ./...`, `go vet ./...`, `go test ./internal/admin/... ./internal/service/...` pass.
- [x] Manual: log in as admin, open `/transactions/e950fza1`, confirm activity rows render gradient avatar, sidebar matches.
- [x] Manual: confirmed timeline now lists 12 events (was rendering "0 events" with `SQLSTATE 22P02` in logs).
- [ ] CI on this PR.

### DIFF
--- a/internal/admin/auth.go
+++ b/internal/admin/auth.go
@@ -126,8 +126,15 @@ func LoginHandler(sm *scs.SessionManager, queries *db.Queries, _ *TemplateRender
 }
 
 // ActorFromSession builds a service.Actor from the session.
+//
+// Prefers the linked users.id over the auth_accounts.id so that actor_id on
+// annotations matches the avatar handler's lookup key (/avatars/{users.id}).
+// Falls back to auth_accounts.id only for unlinked initial admin accounts.
 func ActorFromSession(sm *scs.SessionManager, r *http.Request) service.Actor {
-	id := sm.GetString(r.Context(), sessionKeyAccountID)
+	id := sm.GetString(r.Context(), sessionKeyUserID)
+	if id == "" {
+		id = sm.GetString(r.Context(), sessionKeyAccountID)
+	}
 	name := sm.GetString(r.Context(), sessionKeyAccountUsername)
 	if name == "" {
 		name = "admin"
--- a/internal/db/migrations/20260425131133_backfill_annotation_actor_user_id.sql
+++ b/internal/db/migrations/20260425131133_backfill_annotation_actor_user_id.sql
@@ -1,0 +1,22 @@
+-- +goose Up
+-- Backfill annotations.actor_id for actor_type='user' rows where the recorded
+-- ID is an auth_accounts.id rather than the linked users.id. Before #788 the
+-- admin session built service.Actor from auth_accounts.id, so the avatar
+-- handler at GET /avatars/{id} (which keys on users.id) missed for every
+-- timeline entry written by an admin and fell through to the generated
+-- pattern. ActorFromSession now prefers users.id; this migration repairs
+-- existing rows so older timeline entries also resolve uploaded avatars.
+UPDATE annotations a
+SET actor_id = aa.user_id::text
+FROM auth_accounts aa
+WHERE a.actor_type = 'user'
+  AND a.actor_id IS NOT NULL
+  AND a.actor_id ~ '^[0-9a-fA-F-]{36}$'
+  AND a.actor_id::uuid = aa.id
+  AND aa.user_id IS NOT NULL;
+
+-- +goose Down
+-- One-way data repair. The original auth_accounts.id values are not preserved
+-- per row, so reversal would require re-deriving them from session history,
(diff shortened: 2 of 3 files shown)


canalesb93:
* **#823** <a href="https://app.graphite.com/github/pr/canalesb93/breadbox/823?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/canalesb93/breadbox/823?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #912, opened 2026-04-27
chore: consolidate *string deref helpers into internal/strutil

canalesb93:
## Summary

Six different files defined the same nil-safe pointer-dereference helper under five different names:

| File | Local name |
| ---- | ---------- |
| `internal/admin/synclogs.go` | `stringOrEmpty` |
| `internal/admin/csv_export.go` | `derefStr` |
| `internal/admin/account_links.go` | `derefString` |
| `internal/admin/transactions.go` | `derefOr` (default-aware variant) |
| `internal/service/annotations_enrich.go` | `derefString` |
| `internal/admin/logs.go` | (calls `stringOrEmpty` from synclogs.go) |

All five no-default versions had identical bodies; `derefOr` is the same shape with a fallback. New `internal/strutil` package exposes `Deref(p)` and `DerefOr(p, def)` so future callers have one obvious place to land.

`internal/templates/components/helpers.go` keeps its local `deref()` — that file is intentionally a templ-side mirror of the admin funcMap and short names are part of its ergonomics.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./...` (all unit tests pass)
- [x] New `internal/strutil/strutil_test.go` covers nil, non-nil, and empty-string cases for both helpers

### DIFF
--- a/internal/admin/account_links.go
+++ b/internal/admin/account_links.go
@@ -6,6 +6,7 @@ import (
 
 	"breadbox/internal/app"
 	"breadbox/internal/service"
+	"breadbox/internal/strutil"
 	"breadbox/internal/templates/components"
 	"breadbox/internal/templates/components/pages"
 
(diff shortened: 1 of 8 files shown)


canalesb93:
Superseded by #910 which adds a generic ptrutil.Deref[T] covering the same call sites plus internal/templates/components/helpers.go.

PR #922, opened 2026-04-27
ui(activity): promote tx-detail timeline to main content

canalesb93:
## Summary

Promotes the activity timeline on the transaction-detail page from "another card next to Details and Category" to **the main content section** of the page — closer in spirit to GitHub's PR/issue conversation timeline.

The timeline carries the meaningful chronology of how a transaction has been touched (initial sync → rules → tags → comments), so giving it real estate proportional to that role makes the page feel less like a static record and more like a living conversation.

- Adds a `Variant` field to `components.Timeline` (`""` / `card` keeps the existing chrome for future side-panel callers like sync-log detail and agent-run logs; `prominent` opts into the new treatment).
- Switches the tx-detail caller to `prominent`.
- Preserves the 24px tile / `leading-6` / `ring-4` invariants from `docs/activity-timeline.md` verbatim — the only structural changes are the section chrome (no card, top divider, larger heading) and the rail-mask colours, redirected from `base-100` (card surface) to `base-200` (page surface) via scoped `.bb-timeline-prominent` CSS overrides so individual rows don't have to learn about variants.
- Comment bubbles in the prominent variant pick up a slightly stronger border so they read as discrete messages on the page background.

## Before / After

<table>
  <tr>
    <td>tx detail — before (timeline as card)</td>
    <td>tx detail — after (timeline as main content)</td>
  </tr>
</table>

(The "After" capture happens to include one extra comment vs Before — unrelated dev-DB churn between captures, not a code difference.)

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] Visual check on `/transactions/{id}` with rich activity (Sephora fixture, 9 events spanning sync, rule, tag, category, comments).
- [ ] Reviewer eyeball: empty-activity branch — section heading still renders cleanly without the rail / day-separator chrome (changes the wrapper from `bb-card` to bare section but keeps the same heading + composer composition).

## Notes

- `Timeline` had a single caller before this PR (`txdActivity`) — sync-log detail and agent-run logs are mentioned in `docs/activity-timeline.md` as future surfaces but don't exist yet, so the variant flag is forward-looking rather than a no-op shim.
- Rail offsets and tile sizing are unchanged from the card variant. Anything that broke the alignment in card mode would also break it here.

### DIFF
--- a/input.css
+++ b/input.css
@@ -656,6 +656,33 @@
     word-break: break-word;
   }
 
+  /* ── Activity timeline — prominent (main-content) variant ──
+   * Used by the transaction-detail page to promote the activity timeline
+   * out of card chrome (see internal/templates/components/timeline.templ).
+   * The card-mode 24px tile / leading-6 / ring-4 invariants from
+   * docs/activity-timeline.md are preserved — these overrides only swap
+   * the rail-mask colour from the card surface (base-100) to the page
+   * surface (base-200), so the row tiles still mask the rail seam without
+   * the rows themselves having to learn about variants.
+   *
+   * Rows continue to render with `bg-base-100` and `ring-base-100` because
+   * they're shared with the card variant; here we redirect both at the
+   * variant-scoped level. */
+  .bb-timeline-prominent .bg-base-100 {
+    background-color: var(--color-base-200);
+  }
+  .bb-timeline-prominent .ring-base-100 {
+    --tw-ring-color: var(--color-base-200);
+  }
+
+  /* Comment bubbles in the prominent variant pick up a slightly stronger
+   * border so they read as discrete messages against the (cardless) page
+   * background, matching the GitHub PR-conversation bubble weight. */
+  .bb-timeline-prominent .bb-comment-bubble {
+    background: color-mix(in srgb, var(--color-base-100) 70%, transparent);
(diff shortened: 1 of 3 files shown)


canalesb93:
## v2 — bolder iteration

Promoted the timeline further: scaled tiles + icons up, gave comments a proper GitHub-style bordered card with a header bar and a tail pointing at the avatar, replaced the muted "9 events" suffix with a count chip, and made the rail thicker so it reads against the page surface rather than blending in.

<table>
  <tr><th>Before (card-shaped)</th><th>After v2 (GH-style cards + bigger tiles)</th></tr>
  <tr>
    <td>before — timeline as card</td>
    <td>after v2 — GH-style</td>
  </tr>
</table>

Notes:
- Comment cards target only rows with `.bb-comment-bubble` via `:has()` — system rows and the deleted-comment tombstone keep their compact one-liner shape.
- Composer row picks up the same tail pattern so the conversation reads as one thread.
- Tile / inner-icon / leading scale-up is applied via CSS scoped to `.bb-timeline-prominent` so the shared row markup stays variant-agnostic.
- Heads-up: another agent is reportedly working on the activity log too — I'll resolve any conflicts when this lands behind theirs (or vice versa).

PR #946, opened 2026-04-28
fix(tags): mobile-friendly tag list and actions

canalesb93:
## Summary

Tightens the tag list row layout for narrow viewports: reduces horizontal gaps and chip max-width on mobile, makes the slug `<code>` element block-level with truncation to prevent overflow, moves the transaction count into an inline element beneath the slug on small screens (hidden on `sm+`), and hides the existing desktop count column below `sm`. Also adds `self-start` to the "Add Tag" button header so it doesn't stretch vertically on mobile.

## Screenshots

| Viewport | Before | After |
| --- | --- | --- |
| iPhone (390×844) |  |  |
| Tablet (768×1024) |  |  |
| Desktop (1440×900) |  |  |

## Changes

- `self-start sm:self-auto` on "Add Tag" button — prevents vertical stretch in header flex row on mobile
- Row gap reduced from `gap-3` → `gap-2` on mobile; horizontal padding from `px-4` → `px-3`
- Tag chip max-width reduced from `max-w-[10rem]` → `max-w-[7rem]` on mobile
- Slug `<code>` made `block truncate max-w-full` to prevent long slugs from breaking the layout
- Description line-clamp replaced with `truncate` for single-line overflow control
- Transaction count moved below slug for mobile (`sm:hidden`); desktop count column is now `hidden sm:block`

## Test plan

- [x] Validated at iPhone (390×844), Tablet (768×1024), Desktop (1440×900) via Chrome DevTools MCP

### DIFF
--- a/internal/templates/components/pages/tags.templ
+++ b/internal/templates/components/pages/tags.templ
@@ -21,7 +21,7 @@ templ Tags(p TagsProps) {
 			<h1 class="bb-page-title">Tags</h1>
 			<p class="text-sm text-base-content/50 mt-1 hidden sm:block">Open-ended labels for coordinating work on transactions</p>
 		</div>
-		<a href="/tags/new" class="btn btn-sm btn-primary rounded-xl gap-2">
+		<a href="/tags/new" class="btn btn-sm btn-primary rounded-xl gap-2 self-start sm:self-auto">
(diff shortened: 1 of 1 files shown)


canalesb93:
This needs improvements, instead hide the description while on mobile. Make sure this log doesn't take the full width right now, it looks like it does.

PR #952, opened 2026-04-28
feat(feed): aggressive grouping — sessions, bulk actions, sync-error dedup

canalesb93:
## Summary

Iterates on `/feed` so the page shows **happenings**, not raw annotations. A 200-annotation MCP review session is now one card. Cron-retry storms collapse to one card with an attempt counter. Sync cards render inline transaction samples so you can *see* what came in.

Counter from the dev DB: **690 raw annotations + 50+ failed sync logs** in 3 days collapsed to **14 feed events**.

## Grouping pipeline (`service.ListFeedEvents`)

1. **Hard group by `session_id`** — every annotation in one MCP session collapses into a single **AgentSession** card with a kind-breakdown line ("23 categorised · 12 tagged · 8 commented") and inline transaction samples.
2. **Soft-bucket by `(actor, kind, 5-min)`** — ≥3 same-actor same-kind annotations within a 5-minute window become a **BulkAction** card. Below threshold, only standalone comments survive (every comment is high signal); standalone tag/category set events drop off the home feed (they're still visible on the per-transaction activity timeline).
3. **Sync-error dedup** — repeated cron-retry attempts with the same `(connection_id, error_message)` fold into a single sync card with `RetryCount` populated. The card renders "Failing since 18h · 49 attempts" instead of 49 identical rows.

## Sync cards: inline samples + rule outcomes

Sync cards now show the top 5 actually-new transactions (by absolute amount) inline, with mobile collapsing to 3 visible via `hidden sm:flex`. Rule outcomes from `sync_logs.rule_hits` render under the sample list — "Auto-tag rule applied to 47" — so the rule activity that previously got its own row folds into the parent sync card.

## What got removed

The standalone `tag` / `category` / `rule_batch` event types and their templ renderers. The home feed now renders five card types: **sync**, **report**, **comment** (standalone), **agent_session**, **bulk_action**. Granular tag/category events are still readable on the per-transaction timeline; they just don't dominate the home feed anymore.

## Window

Bounded to the **last 3 days** at the SQL layer (`a.created_at >= NOW() - INTERVAL '3 days'` and equivalent for syncs). Surfaced as a footer note under the rail. Doesn't change with current time the way "last 7 days" would, except for the day labels.

## Decisions

| Knob | Value | Why |
|------|-------|-----|
| Window | 3 days | Bounded so the page never grows unbounded; lines up with "what's happened this weekend". |
| BulkThreshold | 3 | Matches the v1 plan; tuneable. |
| SampleLimit | 5 | 5 desktop / 3 mobile fits comfortably without scrolling. |
| Soft-bucket | 5 min | Long enough to capture a "burst", short enough that a paused-then-resumed session doesn't merge. |

### Desktop — top of feed (1280×900)

The two sync-error cards are the dedup at work: 49 identical Teller-DNS-error retries collapsed into one card, plus a separate 2-attempt cluster for a different error.

Feed v2 — desktop viewport

### Desktop — full scroll

Shows every card type rendered against real data: sync error cluster with retry counts, sync success card with inline transaction samples (WATER SUPPLY COMPANY, MAGNA BANK), standalone comments, a bulk-action card ("admin@example.com updated 4 transactions"), and a 319-transaction sync card with the first 5 inline + "+ 314 more". Footer reads "Showing the last 3 days."

Feed v2 — desktop full page

### Mobile (390×844)

Feed v2 — mobile viewport

<details><summary>Mobile — full scroll</summary>

Feed v2 — mobile full page

</details>

## Test plan

- [ ] On the VM, visit `/feed` and confirm the rail renders with the grouped cards
- [ ] Confirm an MCP session that touches many transactions renders as one AgentSession card
- [ ] Confirm a sync card with N>0 added transactions shows the inline sample strip
- [ ] If a connection has been failing: confirm the sync error card carries "Failing since X · N attempts"
- [ ] Confirm `/` dashboard still renders unchanged

### DIFF
--- a/internal/admin/feed.go
+++ b/internal/admin/feed.go
@@ -1,6 +1,7 @@
 package admin
 
 import (
+	"context"
 	"net/http"
 	"sort"
 	"strings"
@@ -8,24 +9,34 @@ import (
 
 	"breadbox/internal/app"
 	"breadbox/internal/pgconv"
-	"breadbox/internal/ptrutil"
 	"breadbox/internal/service"
 	"breadbox/internal/templates/components/pages"
 )
 
-// FeedHandler serves GET /feed — the activity-style household feed page,
-// designed as a candidate replacement for the legacy stats dashboard. The
-// page renders a chronological, GitHub-style timeline of household events
-// (syncs, agent reports, transaction comments, rule applications, manual
-// recategorizations) with rich cards on the high-signal items and compact
-// rows on the low-signal ones.
+// feedWindowDays is the bounded lookback window for the home Feed page.
+// Events older than this drop off; the cap ensures the page never grows
+// unbounded and that the day buckets stay interpretable. Tuned with the
+// product owner — three days lines up with "what's happened this weekend".
+const feedWindowDays = 3
+
+// FeedHandler serves GET /feed — the activity-style household feed page.
+//
+// The aggregation pipeline is:
+//
+//  1. service.ListFeedEvents returns already-grouped sync / agent_session /
+//     bulk_action / comment events for the last `feedWindowDays`.
+//  2. Reports + connection alerts are pulled here (unrelated to grouping).
+//  3. We project FeedEvents onto the templ-side `pages.FeedItem` shape,
+//     resolve tag/category display names for bulk-action subjects, and
+//     bucket the merged stream by day for the rail's day separators.
 func FeedHandler(a *app.App, svc *service.Service, tr *TemplateRenderer) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
 		ctx := r.Context()
 		now := time.Now()
+		window := time.Duration(feedWindowDays) * 24 * time.Hour
 
-		// Build display lookups so enriched annotation summaries render the
-		// human-friendly category and tag names instead of raw slugs.
+		// Display lookups so bulk-action subjects render with friendly tag
+		// and category display names instead of raw slugs.
 		categoryTree, err := svc.ListCategories(ctx)
 		if err != nil {
 			a.Logger.Error("feed: list categories", "error", err)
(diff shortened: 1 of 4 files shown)


canalesb93:
Iterated on review feedback:

- **Routed every row through the existing Timeline primitives** — `TimelineSystemRowCustomTile` for sync / report / agent_session / bulk_action, `TimelineCommentRowRaw` for comments. The bespoke `bb-card`-per-event scaffolding is gone. Rows now share the rail chrome, 24px tile geometry, and `text-xs / leading-6` body container with the per-transaction activity log on `/transactions/{id}`, so a comment posted from the per-tx composer renders byte-equivalent here.
- **Show user display names** — annotation query now joins through `auth_accounts → users` (same shape as `ListAnnotationsWithActorByTransaction`) so the actor surfaces as `Ricardo Canales` instead of `admin@example.com`. Avatar cache-busting via `users.updated_at` comes from the same join, removing the separate hydrate pass.
- **Consolidated transaction reference** — single shared `feedTxRefRow` / `feedTxRefList` pair is the only place the feed renders a transaction. Sync samples, agent-session samples, and bulk-action samples all flow through it. Easy to upgrade the row visual without touching every event type.
- **Markdown parity** — comment bubbles use `.bb-comment-bubble` with `data-markdown`, picked up by the shared `markdown.js` scanner (same marked + DOMPurify versions / SRI as per-tx page).

Yes — answer to "are you reusing existing timeline components?" — for the row chrome, yes (`Timeline`, `TimelineDay`); for individual event rows, *now* yes (`TimelineSystemRowCustomTile` / `TimelineCommentRowRaw`). Before this commit it was custom `<li>` blocks, which is what produced the "too many card styles" feel.

### After (compact GitHub-style annotations)

Desktop viewport:

Feed v3 — desktop viewport

Desktop full scroll — note rendered comment bodies, "Ricardo Canales" display name throughout, and consistent rail across sync errors, sync success, comments, bulk-action card, and the 319-tx mega-sync at the bottom:

Feed v3 — desktop full

Mobile:

Feed v3 — mobile viewport

<details><summary>Mobile full scroll</summary>

Feed v3 — mobile full

</details>

PR #983, opened 2026-04-29
feat(tx): swap agent-reviewed bot icon for live comment count

canalesb93:
## Summary

- Removes the \`bot\` indicator that lit up whenever an agent had ever set a category on a row (driven by an EXISTS over \`category_set\` annotations with \`actor_type='agent'\`). It surfaced AI activity but missed every human-only comment thread, and once an agent had ever touched the row the icon stayed forever even after a human override.
- Replaces it with a \`messages-square\` icon + count, styled like the existing tag-count badge — visible at a glance regardless of who left the comments. Tooltip is "N comment(s)" via the shared \`pluralS\` helper.
- Counts only live comments: \`SELECT COUNT(*) FROM annotations WHERE kind='comment' AND deleted_at IS NULL\`. Soft-delete tombstones (PR adding \`deleted_at\` on annotations) are excluded so the badge tracks real, current threads.

## Implementation

- \`AdminTransactionRow.AgentReviewed bool\` → \`CommentCount int\`. Both \`ListTransactionsAdmin\` queries (paginated list + by-IDs lookup) and the bench mirror updated in lockstep.
- Three render sites updated: \`tx_row.templ\` desktop meta line + mobile sub-line, and \`tx_row_compact.templ\` (TxRowFeed dense list). All three now hide when \`CommentCount == 0\`.
- Color shifts from \`text-primary/60\` (special-case AI flag) to \`text-base-content/40\` to match the tag-count chip — comment count is generic metadata, not an AI marker.

## Test plan

- [x] \`go build ./...\` clean
- [x] \`go vet ./...\` clean
- [x] \`go test ./internal/templates/... ./internal/service/...\` (unit) passes — including the templ scripts lint
- [ ] Browse \`/transactions\` and confirm rows with comments show the \`messages-square\` + count, rows without comments stay clean. (Skipped in this session: Chrome MCP profile was locked by another session and I didn't want to clobber an active browser.)
- [ ] Confirm soft-deleted comments don't inflate the count

### DIFF
--- a/internal/service/transactions.go
+++ b/internal/service/transactions.go
@@ -670,10 +670,10 @@ func (s *Service) ListTransactionsAdmin(ctx context.Context, params AdminTransac
 		"t.date, t.provider_name, t.provider_merchant_name, t.amount, t.iso_currency_code, " +
 		"t.category_id, c.display_name AS cat_display_name, c.slug AS cat_slug, c.icon AS cat_icon, COALESCE(c.color, pc.color) AS cat_color, " +
 		"t.category_override, t.pending, " +
-		// "Agent reviewed" is inferred from a category_set annotation authored
-		// by an agent. "Pending review" is the presence of the needs-review
-		// tag.
-		"EXISTS(SELECT 1 FROM annotations ann WHERE ann.transaction_id = t.id AND ann.kind = 'category_set' AND ann.actor_type = 'agent') AS agent_reviewed, " +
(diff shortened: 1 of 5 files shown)


canalesb93:
### Merge activity

* **Apr 29, 6:58 AM UTC**: @canalesb93 merged this pull request with [Graphite](https://app.graphite.com/github/pr/canalesb93/breadbox/983).

PR #986, opened 2026-04-29
fix(ui): render feed/timeline day buckets and timestamps in the viewer's timezone

canalesb93:
## Summary

Wall-clock-sensitive surfaces (home /feed timeline, hero "Events today" counters, per-transaction activity log, and the 14-day sync-health map on connection detail) all bucketed against the **server's** local clock. That's fine on a developer laptop but the production VM runs UTC, so a transaction commented on at 11:30 PM Pacific surfaced as "Apr 29 6:34 AM" with a "Today" label sitting on tomorrow's bucket.

This PR plumbs the viewer's IANA timezone through the request, so day grouping and absolute-time tooltips render against the browser's wall clock instead of the server's.

## How

- **`internal/templates/layout/base.html`** — small JS shim sets a `bb_tz` cookie from `Intl.DateTimeFormat().resolvedOptions().timeZone` on every load. Re-set on each render so DST shifts and travel propagate without manual intervention.
- **`internal/admin/timezone.go`** (new) — `UserLocation(r)` reads the cookie, validates via `time.LoadLocation`, falls back to `time.Local`. Tested with valid/empty/garbage zones plus the nil-request guard.
- **`internal/admin/feed.go`** — anchors `now := time.Now().In(loc)` once at the top so `startOfDay` (hero counters), `groupFeedByDay`, and `friendlyDayLabel` all share the viewer's wall clock.
- **`internal/admin/transactions.go`** — same fix on the per-tx detail handler and the timeline rows endpoint; `groupActivityByDay` now takes an explicit `*time.Location`.
- **`internal/admin/connections.go`** — the 14-day sync-health tile map (and the "last successful sync" formatted string) anchor to the viewer's TZ.
- **Templ helpers** — `feedFormatTimestamp(s, now)` and `formatTimelineTimestamp(s, now)` now derive the format location from `now.Location()` instead of `t.Local()`. The shared `timeline.templ` already had `now` in scope at every call site.

## Validation

Captured against the local dev server with `bb_tz=America/Los_Angeles`. The "Spending up 23%…" report row at "27 minutes ago" now renders with `Apr 28, 2026 11:35 PM` in the tooltip (`title=` attribute) — the viewer's wall clock — instead of the previous `Apr 29 6:34 AM` (UTC). The day separator below it correctly labels Apr 28 events as "YESTERDAY" rather than rolling them into "Today".

/feed — viewer-TZ day buckets and tooltips

## Caveat

The very first request from a brand-new session lands without the cookie set yet, so it'll render in the server's TZ for that one paint. The cookie is set during that response, so any reload — or any subsequent navigation — flips to the viewer's clock. Acceptable; the alternative would be a redirect dance for a one-paint cosmetic.

## Out of scope

- The transactions list `relativeDate` helper used by `tx_row_compact` still bucket calendar-only transaction dates against `time.Now()`. Threading loc through the html/template funcMap is a separate refactor — calendar dates have no time component, so the failure mode is different from the timestamp-rendering bug fixed here.

## Test plan

- [x] `go build ./... && go vet ./... && go test ./...` clean
- [x] New `TestUserLocation` covers cookie/empty/invalid/nil-request paths
- [x] Manual browser validation against local dev server: cookie set, tooltip renders viewer-TZ, day separator buckets correctly
- [ ] Verify on the prod VM (UTC) after deploy


canalesb93:
### Merge activity

* **Apr 29, 7:08 AM UTC**: @canalesb93 merged this pull request with [Graphite](https://app.graphite.com/github/pr/canalesb93/breadbox/986).

PR #1004, opened 2026-05-01
v2: ship navigable SPA shell at /v2/ alongside admin UI

canalesb93:
Phase 0 of the v2 frontend plan. Lands a clean SPA shell at \`/v2/\` mounted via \`embed.FS\`, with the existing admin UI completely unchanged at \`/\`.

## What's here

**Stack** — React 18 + TypeScript 5 + Vite 5 + Bun + TanStack (Router/Query) + Tailwind v4 + shadcn/ui (\`new-york\` style). \`sidebar-07\` block adapted to a 4-bucket IA: Money, Setup, Automation, Admin. 14 placeholder pages — the home page renders three intro cards; everything else shows a "Coming soon" placeholder until each gets its own PR.

**Architecture**
- \`web/embed.go\` (package \`webui\`) embeds \`web/dist/\` via \`//go:embed all:dist\`.
- \`/v2/*\` serves the SPA bundle with SPA fallback (unknown extensionless paths return \`index.html\` so the client router resolves them).
- \`/v2/assets/*\` gets long \`Cache-Control\` (Vite content-hashes filenames).
- Bundle missing? Handler serves an inline stub explaining how to build it. \`web/dist/.gitkeep\` is committed; the rest of \`dist/\` is gitignored — real builds populate it.

**Auth & API split**
- \`/api/v1/*\` unchanged (public — API key OR session).
- \`/web/v1/*\` new SPA-internal namespace, **session cookie required**, never accepts API keys, returns JSON 401 (not HTML redirect) so the SPA can detect and route itself. Zero stability promise; never appears in public docs.
- \`/web/v1/me\` returns \`{account_id, username, role}\` from the session.
- New \`SessionUsername\` helper in \`internal/admin/auth.go\`.

**Build pipeline**
- Makefile: new \`web\`, \`web-install\`, \`web-dev\` targets; \`make build\` now depends on \`web\`.
- Dockerfile: installs bun in builder stage, runs SPA build before \`go build\`.
- CI test job: setup-bun + bun install + bun run build before \`go test\`.
- CI release job: same, before cross-platform binary builds.

Self-hoster experience preserved: still one binary, one port, one \`docker compose pull\`. Bun is a build-time dependency only.

## How to test

\`\`\`sh
make build
./breadbox serve
# visit http://localhost:8080/v2/
\`\`\`

Or for HMR development:
\`\`\`sh
# terminal 1
make dev
# terminal 2
make web-dev
# visit http://localhost:5173/v2/
\`\`\`

## Test plan

[...]

PR #1006, opened 2026-05-01
fix: switch v2 SPA build to oven/bun image (unbreaks main)

canalesb93:
**Hotfix — main is currently broken.** The push-to-main after #1004 merged triggered Build & Push, which failed at the bun install step:

\`\`\`
/bin/sh: curl: not found
/bin/sh: /root/.bun/bin/bun: not found
process \"... && curl -fsSL https://bun.sh/install | bash && ...\" did not complete successfully: exit code: 127
\`\`\`

Two issues with the previous approach:
1. \`curl\` isn't in the base alpine image (\`apk add bash unzip\` didn't include it)
2. Even with curl, bun's default Linux binary is glibc-only — alpine ships musl, would need extra detection in the install script

## Fix

Switch to a pinned \`oven/bun:1\` stage that builds \`web/dist/\`, then \`COPY\` the output into the Go builder. Stays on \`BUILDPLATFORM\` (no QEMU emulation), removes the \`apk add\` + bun installer + bun runtime libs from the Go builder layer entirely.

\`\`\`dockerfile
FROM --platform=\$BUILDPLATFORM oven/bun:1 AS web-builder
WORKDIR /web
COPY web/package.json web/bun.lock ./
RUN bun install --frozen-lockfile
COPY web/ ./
RUN bun run build

FROM --platform=\$BUILDPLATFORM golang:1.25-alpine AS builder
...
COPY --from=web-builder /web/dist/ ./web/dist/
\`\`\`

Verified locally: \`docker build --target builder\` completes — ~6s for the SPA, ~13s for the Go binary. Deploy to \`breadbox.exe.xyz\` will fire as soon as this lands on main.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,14 @@
-# Stage 1: Build (runs natively on the build host, cross-compiles Go)
+# Stage 1: Build the v2 SPA bundle. Pinned bun image, runs natively on
+# BUILDPLATFORM (not under QEMU). Output (web/dist/) is copied into the Go
+# builder below.
+FROM --platform=$BUILDPLATFORM oven/bun:1 AS web-builder
+WORKDIR /web
+COPY web/package.json web/bun.lock ./
+RUN bun install --frozen-lockfile
+COPY web/ ./
+RUN bun run build
+
+# Stage 2: Build Go binary (runs natively on the build host, cross-compiles Go)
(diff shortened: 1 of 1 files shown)


canalesb93:
### Merge activity

* **May 1, 6:26 AM UTC**: @canalesb93 merged this pull request with [Graphite](https://app.graphite.com/github/pr/canalesb93/breadbox/1006).

PR #1069, opened 2026-05-14
v2: page-zero hardening before Phase 1 pages

canalesb93:
Hardens the v2 SPA foundation before Phase 1 page-building — driven by a two-subagent review of everything merged so far.

## Why

Before building the ~8 real data pages, two reviewers (code-quality + architecture) audited the foundation. They surfaced two real bugs, two latent problems that would bite the first typed-search-param page, and a set of shared primitives the conventions doc references but that didn't actually exist. All addressed here so Home → Transactions build on solid ground.

## Fixes

- **`useShortcut` effect re-fire** (`lib/shortcuts.ts`) — both `keys` (inline array literal) and `handler` (inline arrow) are referentially unstable and were in the effect's dep array, so the global `keydown` listener was torn down + re-added — and the registry churned — on *every render* of the calling component. Now depends on a stable joined-key string and reads the handler through a ref.
- **Modal search-params vs typed routes** (`lib/modals.ts` + `main.tsx`) — added `baseSearchSchema` (`m`/`ms`) applied to the **root route**, so the modal params survive once a page declares its own `validateSearch`. Without it, Transactions' filter schema would strip `m`/`ms` and overlays would stop opening on that page.
- **Dynamic placeholder routes → explicit override map** (`main.tsx`) — replaced the `NAV_LEAVES.flatMap` generator with a `PAGE_COMPONENTS` override map. Shipping a real page is now one entry that *replaces* the placeholder — no way to silently shadow a real route by forgetting to remove a generated one.
- **`login.tsx`** — cast the redirect target past the typed-router `to` constraint (it's an arbitrary captured pathname, not a statically-known route).

## New shared primitives

- **`DataTable`** — generic wrapper over `@tanstack/react-table` + the shadcn `Table` primitive: built-in loading skeletons, empty/error slots, controlled sorting lifted to the route. Every list page uses this; no hand-rolled `<table>`.
- **`EmptyState`** — zero-data state (icon + title + description + optional CTA), distinct from `routes/placeholder.tsx` which marks unbuilt pages.
- **`PageHeader`** — content-level title/description/actions header, rendered inside `<main>` below the shell breadcrumb.
- **`lib/pagination.ts`** — cursor-pagination helpers (`nextCursor`, `flattenPages`) for the `/api/v1/*` resource-keyed list envelope.
- **`lib/mutation-toast.ts`** — `withMutationToast` wraps the standard `try/catch` + `toast.success`/`toast.error(ApiError.message)` pattern; `account-section.tsx` migrated as the reference call site.

## Test plan

- [x] `tsc -b` + `vite build` + `go build` (default & `-tags=headless`) + `go vet` all pass.
- [x] Browser smoke: login → nav → ⌘K palette → `?` shortcut sheet (both shortcuts registered) → Settings modal at `?m=settings` → placeholder route renders. No regressions.
- [ ] `DataTable` / `EmptyState` / `PageHeader` exercised for real when the Home and Transactions pages land on top of this.

### DIFF
--- a/web/bun.lock
+++ b/web/bun.lock
@@ -14,6 +14,7 @@
         "@radix-ui/react-tooltip": "^1.1.4",
         "@tanstack/react-query": "^5.59.0",
         "@tanstack/react-router": "^1.58.0",
+        "@tanstack/react-table": "^8.21.3",
         "class-variance-authority": "^0.7.1",
         "clsx": "^2.1.1",
         "cmdk": "^1.1.1",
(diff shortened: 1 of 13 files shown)


canalesb93:
* **#1069** <a href="https://app.graphite.com/github/pr/canalesb93/breadbox/1069?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/canalesb93/breadbox/1069?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

canalesb93:
### Merge activity

* **May 14, 4:41 AM UTC**: @canalesb93 merged this pull request with [Graphite](https://app.graphite.com/github/pr/canalesb93/breadbox/1069).

PR #1070, opened 2026-05-14
api: accept the v2 session cookie on /api/v1/*

canalesb93:
Makes `/api/v1/*` accept the v2 dashboard session cookie — so the SPA can read public resources without holding an API key.

## Why

`v2-frontend.md` always claimed `/api/v1/*` was "API key OR session, used by SPA". It wasn't — the whole `/api/v1` block was gated by `mw.APIKeyAuth` (API key / OAuth Bearer only). The v2 SPA authenticates purely by session cookie, so it literally could not call `/api/v1/transactions`, `/categories`, `/connections`, etc. This closes that gap before the first page that needs it.

## How

`sessionOrAPIKeyAuth` (`internal/api/auth_session.go`) wraps the existing `APIKeyAuth` middleware:

- A valid session is translated into a **synthetic `db.ApiKey`** (`ActorType: "user"`), so every downstream `RequireWriteScope` check and actor-attribution path keeps working unchanged — there is no second auth code path to maintain.
- **Scope is role-derived:** `auth_accounts.role` of `admin`/`editor` → `full_access`; `viewer` (or unknown) → `read_only`. Mirrors `admin.IsEditor`.
- **CSRF:** a session-authed *unsafe* method (POST/PUT/PATCH/DELETE) must pass a SameSite=Lax + `Origin`/`Referer` host check — the same guard `/web/v1/*` uses. `/api/v1/*` never had cookie auth before, so writes need it. Pure API-key clients send no cookie and skip the whole branch.
- No session → falls through to the existing API-key / Bearer path, byte-for-byte unchanged.

The `scs.SessionManager` is now created once near the top of `NewRouter` and shared by `/api/v1` and the dashboard mount. It's `nil` under `--no-dashboard` and `-tags=headless` (no cookie clients) — `/api/v1` then behaves exactly as before. Headless stubs added for the `admin.Session*` helpers + role constants so the headless build still compiles.

## Test plan

- [x] `go build` / `go vet` pass for default, `-tags=headless`, and `-tags=lite`.
- [x] Unit tests cover the pure logic: role→scope mapping, unsafe-method classification, same-origin (origin/referer/precedence/port).
- [x] End-to-end: validated by the stacked Transactions PR — the SPA fetches `/api/v1/transactions` with only its session cookie and renders real data.

### DIFF
--- a/.claude/rules/api.md
+++ b/.claude/rules/api.md
@@ -29,6 +29,15 @@ All error responses use:
 - `scope` column: `full_access` or `read_only`. `middleware.RequireWriteScope()` blocks read-only keys from write endpoints.
 - Full key record exposed to handlers via `middleware.SetAPIKey()` / `GetAPIKey()` for actor attribution.
 
+### Session cookie on `/api/v1/*` (the v2 SPA)
+
+`/api/v1/*` also accepts the v2 dashboard **session cookie** — that's how the v2 SPA reads public resources without holding an API key. `internal/api/auth_session.go` (`sessionOrAPIKeyAuth`) wraps the API-key/Bearer middleware:
+
+- A valid session is translated into a **synthetic `db.ApiKey`** (`ActorType: "user"`), so every downstream `RequireWriteScope` / actor-attribution path works unchanged — there is no second auth code path.
+- **Scope is role-derived:** `auth_accounts.role` of `admin` or `editor` → `full_access`; `viewer` (or unknown) → `read_only`. Mirrors `admin.IsEditor`.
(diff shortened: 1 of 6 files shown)


canalesb93:
* **#1071** <a href="https://app.graphite.com/github/pr/canalesb93/breadbox/1071?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1070** <a href="https://app.graphite.com/github/pr/canalesb93/breadbox/1070?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/canalesb93/breadbox/1070?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

canalesb93:
### Merge activity

* **May 14, 5:14 AM UTC**: @canalesb93 merged this pull request with [Graphite](https://app.graphite.com/github/pr/canalesb93/breadbox/1070).

PR #1097, opened 2026-05-15
v2: bulk multi-select on connections list (stack 6/6)

canalesb93:
Final PR (6/6) in the `stack/v2-connections` series. Adds row checkboxes + a sticky selection action bar to `/v2/connections` so households with many institutions can sync, pause, resume, or disconnect connections in a single gesture.

## What's in this PR

- **`features/connections/selection-action-bar.tsx`** — new sticky-bottom pill mirroring the transactions page bar (same `bg-popover` rounded shell, `KbdTooltip` Esc affordance, separator slots). Buttons: **Sync**, **Pause**, **Resume**, **Disconnect…** (destructive), and **Clear** (the X). Each button is gated independently on the slice of the selection it can act on:
  - **Sync** → connections with `provider !== "csv" && status === "active"` (CSV / non-active rows are silently skipped).
  - **Pause** → only `status === "active"` rows.
  - **Resume** → anything that isn't already `active` or `disconnected` (covers paused-active rows and `pending_reauth`/`error` rows that may have been paused alongside).
  - **Disconnect** → applies to every selected row, including CSV. The existing `DELETE /api/v1/connections/{id}` already soft-deletes the connection's transactions for every provider — bulk just fans the same call out.
- **`features/connections/connection-row.tsx`** — accepts new optional `selected` + `onSelectChange` props. When `onSelectChange` is provided, a leading `<Checkbox>` is rendered inside a label with `e.stopPropagation()` on click, so checking the box never navigates to the detail page. Selected rows get a subtle `border-primary/60 ring-primary/20 ring-2` highlight. Per-row sync button + ⋯ menu stay live in select mode (matching the transactions page UX).
- **`routes/connections.tsx`** —
  - Selection state is a `useState<Set<string>>` of UUIDs (Connection.id, not short_id), lifted here so the action bar reads the union and rows fire callbacks back up. Ephemeral by design: not stored in URL search.
  - Header strip above the list shows a tri-state select-all checkbox (`checked` / `indeterminate` / `unchecked`) with a clickable `Select all (N)` / `N selected` label. Selects every visible row after the family-tab filter, not the entire dataset.
  - Switching family tabs clears selection (otherwise hidden rows would be carried along silently).
  - Esc clears selection via `useShortcut("escape", …, { group: "Connections" })`.
[...]

PR #1095, opened 2026-05-15
v2: re-auth sheet (stack 5/6)

canalesb93:
Fifth PR in the `stack/v2-connections` series. Replaces the placeholder `?reauth=$id` Sheet (introduced in PR-01) with the real re-authentication flow inside the v2 SPA.

## What's in this PR

- **`features/connections/reauth-sheet.tsx`** — the new Sheet. Two visible states:
  - **Confirm** — institution name + the connection's last `error_message` (falls back to "The login for this connection has expired…"), one paragraph of context, `[Cancel] [Reconnect]`.
  - **In-flight** — spinner + "Opening Plaid…" / "Opening Teller…" / "Reconnecting…" while the SDK runs in update / reconnection mode. The Sheet stays open until the user dismisses it or the completion call resolves.
- **Provider routing** — picked from the connection record (the Sheet fetches its own `useConnection(shortId)` so callers pass only the short_id):
  - **Plaid** → POST `/api/v1/connections/{id}/reauth` mints an update-mode link token. Reuses `PlaidLinkButton` unchanged (per PR-03's report) — Plaid Link runs in update mode automatically when given an update-mode token.
  - **Teller** → same endpoint; Teller's `CreateReauthSession` returns the existing `enrollment_id` as the "link token" (see `internal/provider/teller/reauth.go`). `TellerConnectButton` gains a thin `enrollmentId` prop that, when set, switches Teller Connect into reconnection mode (the SDK key is `enrollmentId` in `TellerConnect.setup`). No new endpoint, no new wrapper file.
  - **CSV** — defensive-render an "N/A" message rather than crash. The Connect/Detail pages don't surface a re-auth affordance for CSV, so this only fires when the URL is hand-crafted.
- **Completion** — both providers call POST `/api/v1/connections/{id}/reauth-complete` from `onSuccess`. Toast `Reconnected $institutionName.`, invalidate `["connections"]` + `["connection"]` + `["sync-logs"]`, close the Sheet.
- **Two new query hooks** in `api/queries/connections.ts`: `useReauthStart()` (returns the existing `LinkSession` shape) and `useReauthComplete()`.
- **Wiring** — `routes/connections.tsx` and `routes/connection-detail.tsx` swap the `ComingSoonSheet` instance for `<ReauthSheet open={!!search.reauth} connectionShortId={search.reauth} … />`. The detail page reads `?reauth=` from its own search schema (added by PR-02), so the same Sheet opens from the list, the detail-page banner, and the ⋯ menu.
[...]

PR #1094, opened 2026-05-15
v2: CSV import in connect-bank sheet (stack 4/6)

canalesb93:
Fourth PR in the `stack/v2-connections` series. Adds a CSV branch to the Connect-bank Sheet from PR-03 and an "Import more" button on the connection-detail page so admins can append rows to an existing CSV connection.

## What's in this PR

- **`features/connections/csv-import-form.tsx`** — the upload → mapping → import form. Two stages inside the same Sheet:
  - **Drop / browse** — `<label>`-wrapped `<input type="file" accept=".csv">` with drag-and-drop styling. POSTs to `/api/v1/connections/csv/preview` (multipart) on file selection. 10MB client-side cap with a friendly toast (backend still allows 50MB; the smaller cap gives faster local feedback).
  - **Map columns + import** — `<Select>` per recognised field, pre-filled from the preview's `inferred_mapping`. Required fields (date, description, amount or debit/credit) are validated client-side before the import button enables. Includes a debit-vs-credit toggle and a positive-vs-negative sign convention; both honour the auto-detected template defaults. Renders a 10-row preview table below.
- **`features/connections/connect-bank-sheet.tsx`** — adds a `csv` arm to the discriminated `Stage` union. Picking CSV in the picker + Continue swaps the Sheet body to `CsvImportForm`. New optional `appendToConnectionId` prop lets the same Sheet skip the picker and pre-target an existing connection in append mode (used by the detail page below). On success, toast and navigate to the new connection's detail page; in append mode, just close.
- **`routes/connection-detail.tsx`** — adds an `import_csv` search param and an "Import more" header button for CSV connections (`provider === "csv"`). The button opens the same `ConnectBankSheet` inline, pre-targeted to this connection. URL-driven so the sheet is shareable / back-button-friendly, mirroring the existing `?reauth=` pattern.
- **`features/connections/provider-picker.tsx`** — CSV is now in the default `providers` list (was Plaid + Teller only). The Connect-bank Sheet always marks CSV as enabled — the importer is in-binary, not a credentialled provider — so the previous "Not configured" placeholder is gone.
- **Queries** — `useCsvPreview()` and `useCsvImport()` in `api/queries/connections.ts`, both `multipart/form-data`. Import invalidates `connections`, `connection`, `accounts`, and `transactions`.
- **`api/types.ts`** — adds `CsvPreviewResult`, `CsvImportResult`, and a `CsvColumnKey` union mirroring the column names the backend recognises.

## `client.ts` change worth flagging

[...]

PR #1093, opened 2026-05-15
v2: connect-bank sheet — Plaid + Teller (stack 3/6)

canalesb93:
Third PR in the `stack/v2-connections` series. Replaces the `?action=connect` placeholder Sheet with a real Plaid + Teller Connect flow.

## What's in this PR

- **`features/connections/connect-bank-sheet.tsx`** — two-step Sheet replacing `ComingSoonSheet` for `?action=connect`. Step 1 is provider + family-member; step 2 hands off to the matching SDK and persists the connection on success.
- **`features/connections/provider-picker.tsx`** — reusable card list. Takes `enabledProviders` so providers without server-side credentials render disabled (`Not configured` hint). Already future-proofed for CSV (PR-04) via the optional `providers` prop. Carries a sandbox specimen.
- **`features/connections/plaid-link-button.tsx`** — wraps `usePlaidLink` from the official `react-plaid-link`. Renders nothing visible; auto-opens Plaid Link on `ready`. Same shape will service PR-05 (re-auth, update mode).
- **`features/connections/teller-connect-button.tsx`** — script-tag launcher for `https://cdn.teller.io/connect/connect.js` (Teller doesn't ship a React wrapper). Same `autoOpen` + `onSuccess`/`onExit`/`onFailure` shape as the Plaid wrapper so the Sheet branches on `provider` only.
- **Queries** — extended `api/queries/connections.ts` with `useProviders()`, `useProviderLinkSession()` (mutation, since each token is one-shot), and `useCreateConnection()` (POSTs to the generic dispatch endpoint). Mirrors response/credential shapes added to `api/types.ts`.

## What's deliberately NOT here

- **CSV import** → PR-04 (slots into `ProviderPicker` via the `providers` prop).
- **Re-authentication** → PR-05 (re-uses `PlaidLinkButton` in update mode against an existing connection's link token).
- **Bulk select** → PR-06.
- **No new backend endpoints.** This PR consumes the generic `/api/v1/providers`, `/api/v1/providers/{name}/link-session`, and `/api/v1/connections` surfaces shipped earlier.
- **No global state library, no fetch outside `client.ts`.** Plaid + Teller SDK loads happen via the official hook / a one-shot script-tag injector (cached in module scope).

## Empty-config behaviour

[...]

PR #1098, opened 2026-05-15
v2: simplify-pass over connections stack (stack 7/7)

canalesb93:
Cleanup pass over PRs [#1089](https://github.com/canalesb93/breadbox/pull/1089), [#1090](https://github.com/canalesb93/breadbox/pull/1090), [#1093](https://github.com/canalesb93/breadbox/pull/1093), [#1094](https://github.com/canalesb93/breadbox/pull/1094), [#1095](https://github.com/canalesb93/breadbox/pull/1095), and [#1097](https://github.com/canalesb93/breadbox/pull/1097). Two real bug fixes plus dead-code/comment trimming. No new features.

## What changed

- **Fix infinite-loop on `/v2/connections`.** `ReauthSheet` had a `useEffect` whose dep array included the `useReauthStart()` and `useReauthComplete()` mutation result objects. TanStack Query rebuilds those wrapper objects on every render (the methods inside are stable), so the effect re-fired every render and called `mutation.reset()` — which itself triggers a query-store re-render. Result: the React "Maximum update depth exceeded" warning the PR-06 agent flagged. Fix is to depend only on `[open]` and read the mutations through refs.
- **Fix bad route param in the post-create navigate.** `ConnectBankSheet` and the CSV success handler navigated to `to: "/connections/$short_id"` with `params.short_id`, but the registered route is `path: "/connections/$id"`. The mismatch silently failed. Now uses `to: "/connections/$id"` with `params.id`.
- **Defer the per-connection `useSyncLogs` fetch** until the connection is loaded. Previously fired a global-feed request on first render, then immediately discarded it.
- **Drop the misleading "Resume syncing" item from the connection-row menu.** The list endpoint doesn't return `paused`, so the row can't tell paused-active rows from active rows. The dropped item was wired to fire only on `error` / `pending_reauth` rows, which actually need re-auth, not a paused-flag flip. Resume lives on the detail page (where `paused` is visible) and on the bulk action bar.
- **Drop the unused `paused?: boolean` prop on `ConnectionRowProps`.** Added in PR-01 for an optimistic-pause UI that never landed.
- **Sweep PR-NN comment references** out of the touched files (`PR-01` / `PR-02` / `PR-03` / `PR-05` mentions in `connections.tsx`, `connect-bank-sheet.tsx`, `plaid-link-button.tsx`, `teller-connect-button.tsx`, `connection-detail.tsx`).
- **Drop the "(coming soon)" qualifier** on the no-providers Alert copy — CSV is in-binary and always available.
- **Tidy import grouping** in `connection-detail.tsx` (an `Upload` icon was imported on its own trailing line; folded into the main lucide import block).

## What I considered and rejected

- **Generic `<ResourceList>` extraction across `connections.tsx` and (eventual) `accounts.tsx`.** Right shape, but `accounts.tsx` is still a placeholder — extracting now would be inventing the abstraction off one call site. Revisit when `/accounts` lands.
- **Collapsing `countsByUser` + `countsByUserShortId` into a single pass over users.** It's already O(N + M) and reads more clearly as two narrow steps than as one fused loop.
- **Removing the dead `enabledProviders.length === 0` branch in `ConnectBankSheet`** (CSV is always pushed onto the array, so the branch can't fire). Visually inert; removing changes the surface area of an already-validated component for zero behavioural gain. Left as defensive.
- **Replacing the row's `Pause` action with a real toggle.** Would require either fetching the per-connection detail per row or adding `paused` to the public `Connection` shape. Both are out of scope for a simplify pass.

## "Maximum update depth" warning

Reproduced on a fresh load of `/v2/connections` before the fix. After the `ReauthSheet` effect change the warning is gone (verified by re-loading the page with the dev console captured — only event was an unrelated 404 on the login route).

## Validation

- `bun run lint` ✓
- `bun run build` ✓
- `templ generate && go build ./... && go vet ./... && go test ./...` ✓
- Smoke `/v2/connections` and `/v2/connections/$id` in the dev browser, console captured, no React update-depth warning observed.

Screenshot of `/v2/connections` after the cleanup: connections list

Screenshot of the detail page: connection detail

## Stack is now 7/7 complete

Stack: 1089 → 1090 → 1093 → 1094 → 1095 → 1097 → this. Ready for review end-to-end.

### DIFF
--- a/web/src/api/queries/sync-logs.ts
+++ b/web/src/api/queries/sync-logs.ts
@@ -36,12 +36,19 @@ export interface UseSyncLogsParams {
   // Connection UUID (short_id is not accepted by /api/v1/sync/logs).
   connectionId?: string;
   limit?: number;
+  // Defaults true — pass false to defer the fetch until the parent has the
+  // info it needs (e.g. resolving a short_id → UUID first).
+  enabled?: boolean;
 }
 
 // useSyncLogs fetches the most-recent sync history. Pass a connection UUID
 // (not the short_id) — the backend filter only resolves UUIDs. Omitting
 // `connectionId` returns the global feed.
-export function useSyncLogs({ connectionId, limit = 30 }: UseSyncLogsParams) {
+export function useSyncLogs({
+  connectionId,
+  limit = 30,
+  enabled = true,
+}: UseSyncLogsParams) {
   return useQuery({
     queryKey: ["sync-logs", connectionId ?? null, limit],
     queryFn: () => {
@@ -50,7 +57,7 @@ export function useSyncLogs({ connectionId, limit = 30 }: UseSyncLogsParams) {
       qs.set("limit", String(limit));
       return api<SyncLogsPage>(`/api/v1/sync/logs?${qs.toString()}`);
     },
-    enabled: connectionId == null || !!connectionId,
+    enabled,
     staleTime: 30_000,
   });
 }
--- a/web/src/features/connections/connect-bank-sheet.tsx
+++ b/web/src/features/connections/connect-bank-sheet.tsx
@@ -64,7 +64,7 @@ type Stage =
 //      TellerConnectButton) which auto-opens its hosted UI. On success the
 //      launcher hands back the public_token / enrollment payload, we POST
 //      to the generic /api/v1/connections endpoint, toast success, and
-//      navigate to the new connection's detail page (PR-02).
+//      navigate to the new connection's detail page.
 export function ConnectBankSheet({
   open,
   onOpenChange,
@@ -224,8 +224,8 @@ export function ConnectBankSheet({
       toast.success(`Connected ${result.institution_name}.`);
       handleSheetChange(false);
       navigate({
-        to: "/connections/$short_id",
-        params: { short_id: result.connection_id },
+        to: "/connections/$id",
+        params: { id: result.connection_id },
       });
     } catch (err) {
       const msg =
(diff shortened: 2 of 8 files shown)


canalesb93:
* **#1098** <a href="https://app.graphite.com/github/pr/canalesb93/breadbox/1098?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/canalesb93/breadbox/1098?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1097** <a href="https://app.graphite.com/github/pr/canalesb93/breadbox/1097?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1095** <a href="https://app.graphite.com/github/pr/canalesb93/breadbox/1095?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1094** <a href="https://app.graphite.com/github/pr/canalesb93/breadbox/1094?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1093** <a href="https://app.graphite.com/github/pr/canalesb93/breadbox/1093?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1090** <a href="https://app.graphite.com/github/pr/canalesb93/breadbox/1090?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1089** <a href="https://app.graphite.com/github/pr/canalesb93/breadbox/1089?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

canalesb93:
### Merge activity

* **May 15, 11:33 AM UTC**: @canalesb93 merged this pull request with [Graphite](https://app.graphite.com/github/pr/canalesb93/breadbox/1098).

PR #1106, opened 2026-05-16
v2: API keys management pages

canalesb93:
## Summary

Build out the v2 SPA's API keys surface — list page, new-key form, and one-time plaintext disclosure subpage. Drops the user into the same conventions as the rest of v2 (shadcn primitives, RHF + zod, shared `DataTable`, `withMutationToast`).

- **`/v2/api-keys`** — Active / Revoked tabs over the shared `DataTable`, with name/prefix/actor search. Per-row dropdown opens a revoke confirm dialog that wires straight into `useRevokeAPIKey`. Read-only sessions hit the existing `ApiError` path because `GET /api/v1/api-keys` requires write scope.
- **`/v2/api-keys/new`** — shadcn `Form` + RHF + zod. Scope and actor type are `OptionCard`-wrapped `RadioGroup` tiles so the choice reads as a deliberate decision rather than a buried dropdown. Optional actor display name lets you label an integration ("Claude — Ricardo's laptop") without renaming the key itself.
- **`/v2/api-keys/created`** — `Card` with a one-time-copy plaintext input, an `X-API-Key` usage hint, and a one-line "stored as SHA-256 hash" warning. Plaintext lives in `sessionStorage` so a refresh mid-copy isn't fatal; it clears on unmount or on **"I've saved it — done"**.

New shadcn primitive (`radio-group`) added via the CLI plus a sandbox specimen so the component is discoverable in the gallery.

## Screens

**List — Active / Revoked tabs, search, per-row actions** (desktop / tablet / mobile):

API keys list

**New key form** (desktop):

New API key form

**New key form — responsive composite**:

New API key form — responsive

**Key created — one-time plaintext disclosure**:

Key created — one-time disclosure

## Test plan

- [x] `bun run lint` (tsc --noEmit) — clean
- [x] `bun run build` — clean
- [x] Validated `/v2/api-keys`, `/v2/api-keys/new`, `/v2/api-keys/created` in headless Chromium against a locally-built `breadbox serve` on port 8083
- [x] Drove the new-key form end-to-end — list refetches and the new row shows up after returning from `/created`
- [x] Sandbox `RadioGroup` specimen renders in `/v2/sandbox` → Primitives

### DIFF
--- a/web/src/api/queries/api-keys.ts
+++ b/web/src/api/queries/api-keys.ts
@@ -1,0 +1,51 @@
+import { useMutation, useQuery, useQueryClient } from "@tanstack/react-query";
+import { api } from "@/api/client";
+import type {
+  APIKey,
+  APIKeyActorType,
+  APIKeyScope,
+  CreateAPIKeyResult,
+} from "@/api/types";
+
+// useAPIKeys loads the full key catalog. The list endpoint requires a
+// full_access scope (session admins/editors qualify) — read-only sessions
+// will see a 403 surfaced as ApiError.
+export function useAPIKeys() {
+  return useQuery({
+    queryKey: ["api-keys"],
+    queryFn: () => api<APIKey[]>("/api/v1/api-keys"),
+    staleTime: 30_000,
+  });
+}
+
+export interface CreateAPIKeyInput {
+  name: string;
+  scope: APIKeyScope;
+  actor_type: APIKeyActorType;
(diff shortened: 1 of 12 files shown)


canalesb93:
### Merge activity

* **May 16, 6:09 AM UTC**: @canalesb93 merged this pull request with [Graphite](https://app.graphite.com/github/pr/canalesb93/breadbox/1106).

PR #1112, opened 2026-05-16
fix: route Teller link-session through provider so app_id reaches the SPA

canalesb93:
## Summary

- The generic `/api/v1/providers/{name}/link-session` handler returns 204 No Content for any entry with `needsLinkSession:false`. Teller had this flag false, so `TellerProvider.CreateLinkSession` (which returns the configured `app_id` as the link token) was never called.
- The connect sheet then read `window.__TELLER_APP_ID__` (a global nobody sets) and toasted "Teller application ID is not configured on this server" — even though Test connection on `/v2/providers` reported the Teller integration as healthy.

Two-line fix:
- `internal/api/providers_teller.go`: flip `needsLinkSession` to true so the handler invokes the provider.
- `web/src/features/connections/connect-bank-sheet.tsx`: read `session.link_token` instead of the unset window global.

## Test plan

- [x] Reproduced locally — Connect → Teller → Continue threw the toast on `main`.
- [x] After the fix: `POST /api/v1/providers/teller/link-session` returns `200` with `{link_token: "app_..."}`, Teller Connect iframe opens with that `app_id`, and the sheet advances to the in-iframe enrollment screen.
- [x] No change to Plaid path; CSV still skips the link-session call entirely.

### DIFF
--- a/internal/api/providers_teller.go
+++ b/internal/api/providers_teller.go
@@ -29,7 +29,7 @@ type tellerCredentials struct {
 
 var tellerEntry = providerEntry{
 	name:             "teller",
-	needsLinkSession: false, // Teller Connect runs entirely client-side; no init token
+	needsLinkSession: true, // Teller Connect needs the server-configured app_id; returned as link_token
 	capabilities:     []string{"transactions", "balances"},
 	credentialsSchema: map[string]CredentialField{
 		"access_token": {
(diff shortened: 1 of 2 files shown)


canalesb93:
Superseded — the connect-flow half landed in #1114, and the reauth-flow half (the open follow-up that #1114 punted on) ships in #1219.

PR #1143, opened 2026-05-16
design(v2): wire next-themes — Light/Dark/System switcher in NavUser

canalesb93:
## Summary

- Wire `<ThemeProvider>` from `next-themes` in `main.tsx` (class attribute, system default, `breadbox-theme` storage key) so `useTheme()` finally has a provider — Sonner's `useTheme()` call was already in the codebase but reading from no provider, silently falling through to "system".
- Replace the inline OS-follow script in `index.html` with a no-flash bootstrap that respects `localStorage.breadbox-theme` first, then OS preference. OS changes only auto-apply while the saved choice is `system`.
- Add a Theme submenu to `<NavUser>` (Light / Dark / System) using the existing dropdown submenu primitives. Trigger shows the current choice as a muted eyebrow; active option gets a `<Check>`.

Dark-mode tokens were already shipped in `globals.css`; this is the plumbing that lets a user actually pick one. Unlocks dark-mode auditing of the rest of the v2 pages in follow-up iterations.

## Theme switcher (NavUser dropdown → Theme submenu)

switcher-open

## Home page — Light vs Dark

| Light | Dark |
| --- | --- |
| light | dark |

## Notes

- Storage key is namespaced (`breadbox-theme`) to avoid clobbering whatever the classic UI may write to the unprefixed `theme` key.
- Sonner's `useTheme()` was already in the tree but had no provider — toasts will now follow the user's choice instead of always defaulting to system.
- Submenu uses `DropdownMenuPortal` so it floats above the sidebar even when the sidebar collapses.
- Dark-mode home renders cleanly — no obviously-broken tokens — but a full audit across detail / form pages is queued for a follow-up iteration.

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/web/index.html
+++ b/web/index.html
@@ -45,16 +45,34 @@
     <meta name="twitter:image" content="./og-image.svg" />
 
     <script>
-      // Light/dark theme follows the OS — same contract as `BrandHeader`
-      // and the rest of the v2 SPA. Inline + synchronous so the first
-      // paint never flashes the wrong theme.
+      // No-flash theme bootstrap. Inline + synchronous so the first
+      // paint never flashes the wrong theme. Reads the same storage
+      // key (`breadbox-theme`) that `ThemeProvider` writes from
+      // `next-themes` so a user's saved choice survives reloads.
+      // Falls back to the OS preference when nothing is saved or
+      // when the saved value is "system".
       (function () {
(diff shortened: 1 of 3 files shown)


canalesb93:
Superseded — iter 30 (#1142) already shipped a richer ThemeProvider + NavUser theme switcher; this branch landed after that and is redundant. Closing.

PR #1307, opened 2026-05-18
fix(v2): stop SoftBackButton from stretching to full width

canalesb93:
## Summary

- `SoftBackButton` is an `inline-flex` ghost link, but every consumer renders it inside a `flex flex-col` parent (the app `<main>` shell or a `mx-auto flex max-w-{2xl,5xl} flex-col` wrapper around detail/form pages). With the flex default `align-items: stretch`, the link gets blown out to full width — content stays centered (because the button is `justify-center`), so the back-link visually drifts to the middle of the page instead of sitting on the left edge.
- Visible on tag detail, transaction detail, account detail, category detail, connection detail, and the corresponding `*-new` routes. The agents pages worked around it by passing `className="self-start"` per call site — a hint that the primitive itself was wrong.
- Fix at the primitive level: bake `self-start` into `SoftBackButton`'s base classes so every caller is correct by default, and drop the now-redundant overrides from the three agents routes. `self-start` is a no-op in non-flex parents, so this is safe everywhere it currently renders.

back button — before vs after on /v2/tags/:slug

The before frame on the left simulates the pre-fix state by overriding `align-self: stretch` on the rendered link at runtime; the after frame is the page as it ships in this PR.

## Test plan

- [x] `bun run lint` clean
- [x] Visual check at `/v2/tags/meaningless-tag` — link sits at content width (~112px), `align-self: flex-start`, hover surface no longer spans the full 672px container
- [x] Visual check at `/v2/agents/new` — formerly relying on the per-call-site `self-start` override; identical render after dropping the override
- [ ] Spot-check `/v2/transactions/:id`, `/v2/accounts/:id`, `/v2/categories/:id`, `/v2/connections/:id`, `/v2/tags/new`, `/v2/categories/new`, `/v2/api-keys/new`, `/v2/rules/new` on review

### DIFF
--- a/web/src/components/soft-back-button.tsx
+++ b/web/src/components/soft-back-button.tsx
@@ -28,6 +28,13 @@ interface SoftBackButtonProps {
 // No bottom margin — the page layout (<main> or a constrained-width wrapper)
 // is a flex column with gap-5 that owns the rhythm between the back-link and
 // the PageHeader below it. Don't fork the look — change this primitive instead.
+//
+// `self-start` is baked in because every consumer renders this inside a
+// `flex flex-col` parent (the app `<main>` or a `mx-auto flex max-w-* flex-col`
+// wrapper). Without it the flex default `align-items: stretch` blows the
+// inline-flex button out to full width — content stays centered but the
+// hover/focus surface and visual weight read as a banner instead of a link.
+// In non-flex parents `self-start` is a harmless no-op.
 export function SoftBackButton({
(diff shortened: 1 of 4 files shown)


canalesb93:
### Merge activity

* **May 18, 7:03 AM UTC**: @canalesb93 merged this pull request with [Graphite](https://app.graphite.com/github/pr/canalesb93/breadbox/1307).

PR #1311, opened 2026-05-18
feat(admin): add v2_default toggle to land at /v2/ from /

canalesb93:
## Summary

Adds an opt-in app-config toggle so a household can land at the v2 SPA from the v1 admin root (`/`) ahead of the broader v1 retirement. Default behaviour is unchanged.

- **New `v2_default` `app_config` key** — `appconfig.KeyV2Default`. Off by default; when `"true"`, `admin.FeedHandler` returns a 303 to `/v2/` after the existing onboarding gate.
- **Settings → System tab** gains a "Default landing UI" card with a `toggle` checkbox posting to `POST /settings/v2-default` (CSRF-guarded, admin-only, registered alongside `/settings/sync` and `/settings/retention`).
- Onboarding still wins — if `onboarding_dismissed` is false, `/` keeps redirecting to `/getting-started` regardless of `v2_default`, so first-run is untouched.
- Every other v1 admin path (`/transactions`, `/connections`, `/settings/*`, etc.) keeps working; only `/` flips. This is deliberate — v2 doesn't cover every surface yet, and we still want deep links to resolve.

Tracked under #1313 (full v1 admin retirement).

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./internal/admin/... ./internal/appconfig/... ./internal/templates/...`
- [ ] **Browser screenshot of the new toggle is not attached** — this worktree didn't have a dev server set up (no ENCRYPTION_KEY accessible from the sandbox) and the change is small + additive (reuses the `bb-card` / `bb-icon-header` / form pattern from the adjacent `settingsSystemCard`). Reviewer should sanity-check the toggle on the System tab after pulling.
- [ ] Manual: toggle on → visit `/` → land on `/v2/`. Toggle off → `/` renders v1 feed again. First-run (`onboarding_dismissed=false`) still routes to `/getting-started` either way.

### DIFF
--- a/internal/admin/feed.go
+++ b/internal/admin/feed.go
@@ -44,6 +44,14 @@ func FeedHandler(a *app.App, svc *service.Service, tr *TemplateRenderer) http.Ha
 			return
 		}
 
+		// Opt-in: when v2_default is set, the v1 admin root becomes a
+		// redirect to the v2 SPA. Bookmarks and deep links into other v1
+		// pages keep working — only `/` flips.
+		if appconfig.Bool(ctx, a.Queries, appconfig.KeyV2Default, false) {
+			http.Redirect(w, r, "/v2/", http.StatusSeeOther)
+			return
+		}
+
 		// Anchor every wall-clock decision (day buckets, "Today" hero
 		// counters, absolute-time tooltips) to the viewer's browser
 		// timezone via the bb_tz cookie. Falls back to server local when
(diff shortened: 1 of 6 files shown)


canalesb93:
### Merge activity

* **May 18, 7:30 AM UTC**: @canalesb93 merged this pull request with [Graphite](https://app.graphite.com/github/pr/canalesb93/breadbox/1311).

PR #1329, opened 2026-05-18
fix(v2): refresh router + queries on iOS Safari bfcache restore

canalesb93:
## Summary

Fixes user-reported T1: on iOS Safari, swiping from the left edge to go back sometimes shows a blank page and/or freezes mid-swipe. Tapping the toolbar back button works fine — pointing at a bfcache (back-forward cache) restore issue, not a generic routing bug.

## What's happening

When Safari navigates back via swipe, it tries to restore the previous page from bfcache. bfcache stores a full snapshot of the page (DOM + JS state). On restore:

1. `pageshow` fires with `event.persisted === true`.
2. The DOM is reconstituted as-is.
3. JS is **resumed**, not restarted — React state and the TanStack Query cache are exactly as they were minutes ago.

Two interactions specific to this SPA conspire to produce the freeze:

- **No `pageshow` handler exists anywhere in `web/src/`** — so when bfcache restores us, the React tree and TanStack Router state are stale, including any in-flight mutations, pending dialogs, or out-of-date URL state.
- **TanStack Query `staleTime: 30_000`** means queries refetch on focus. A stale-session refetch returning 401 races with Safari's swipe-back restore animation: the layout's 401 handler navigates to `/login` *during* the restore, presenting as a frozen / blank page.

The page is correctly bfcache-eligible (no `beforeunload` / `unload` / `pagehide` handlers, no live WebSocket / EventSource / IndexedDB connections, no `Cache-Control: no-store` on `index.html`). bfcache makes back-navigation feel instant on iOS — we want to keep it. The fix is to handle restoration gracefully.

## The fix

Add a `pageshow` listener in `web/src/main.tsx` after the router + query client are constructed. On `event.persisted === true`:

1. `router.invalidate()` — re-runs active loaders + search-param validation, picking up any URL desync between TanStack Router's snapshot and `window.location`.
2. `queryClient.invalidateQueries()` — marks all queries as stale so they refetch with fresh data. We deliberately do not `queryClient.clear()` — that would drop cached data instantly and flash skeletons, which is jarring after a swipe-back.

Module-scoped because `router` and `queryClient` are module-scoped — no hook needed, no extra render dependency. The change is ~10 LOC plus a long comment explaining the non-obvious "why" so future-us doesn't strip it.

## What's intentionally not changed

- `__root.tsx` 401 handler stays as-is. It's already a `useEffect` keyed on `is401` + a `useMe` query that respects `staleTime` — once the bfcache restore invalidates queries, the 401 path runs from a fresh refetch instead of replaying a stale 401. No visibility-gated debounce needed.
- bfcache itself is not disabled — `no-store` on `index.html`, `beforeunload`, etc. all opt the page out and hurt every back-navigation to "fix" this corner case.
- `staleTime` defaults stay put.
- TanStack Router not migrated.

## How to test

1. Open `/v2/home` on iOS Safari (or any WebKit browser that exposes the swipe-back gesture).
2. Navigate to `/v2/transactions`.
3. Swipe from the left edge to go back.

**Before:** the swipe-back animation freezes or lands on a blank page; only a hard refresh recovers.
**After:** the swipe lands on `/v2/home` with current data and no freeze. The toolbar back button continues to behave correctly.

Desktop browsers don't expose the swipe gesture but you can exercise the same code path by toggling to another tab and back, or using DevTools' Application panel → Back/Forward Cache to force a `pageshow.persisted` event.

No screenshots — this is a behavior fix, not a layout change.

## Test plan

- [x] `cd web && bun run lint` clean
- [x] `cd web && bun run build` clean
- [ ] iOS Safari swipe-back from `/v2/transactions` → `/v2/home` no longer freezes
- [ ] Toolbar back button still works
- [ ] No regression in TanStack Router preload / loader behavior
- [ ] Authenticated 401 path still redirects to `/login` correctly when session expires

### DIFF
--- a/.claude/mobile-sprint-state.md
+++ b/.claude/mobile-sprint-state.md
@@ -11,13 +11,25 @@
 Phase 1 shipped fixes (now in main):
 - Sidebar close-on-tap, popover/select/dropdown collision + width clamps, 44pt tap targets via `pointer-coarse:` pseudo, `viewport-fit=cover` + safe-area, table scroll-shadow + iOS momentum, dvw/dvh viewport units, PageHeader wrap + FormFooter stacking, agent-form CSS-order reshuffle, empty-state width + error pre momentum, transactions filter chip rail, settings tab scroll.
 
-## Workflow (unchanged)
+## Workflow — PROACTIVE MODE
 
-1. Loop wakes → sync branch → merge ready PRs → pick next backlog item.
-2. Spawn ONE subagent in worktree isolation. Brief includes file paths + iOS quirks.
-3. Subagent opens PR against this branch with `img402.dev` screenshots at 375×812 / 768×1024 / 1280×800.
-4. Orchestrator reviews next iteration. Merges when green. Never merge to `main` directly without explicit auth.
-5. When backlog is meaningfully accumulated (3+ merged), open a fresh `sprint → main` bundle PR.
+The loop NEVER reports idle. Every iteration must produce a merge, a dispatch, or a scout-then-dispatch chain.
+
+Priority ladder — try each tier until something dispatches:
+- **T1 User-reported bugs** (highest)
+- **T2 Backlog items in state doc** (P0/P1/P2 including deferred)
+- **T3 Fresh scout on a previously-unaudited surface** (rotate; don't repeat within 5 iterations)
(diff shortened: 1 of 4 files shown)


canalesb93:
Re-opening to trigger CI on the retargeted base (main)

PR #1439, opened 2026-05-24
feat(design): add transaction rows section to sandbox

canalesb93:
## Summary

Adds a **Transaction rows** section to `/design` under the Data display group, covering all three live row variants (`TxRow`, `TxRowCompact`, `TxRowFeed`) plus the building-block primitives that compose them.

- **Full row** (`TxRow`): four sample rows show categorized + tagged + commented, income (uncategorized), pending with clock icon, and a settled uncategorized row.
- **Compact row** (`TxRowCompact`): the same four rows in the dense two-line shape used on the dashboard and rule detail pages.
- **Feed one-liner** (`TxRowFeed`): same four rows in the single-line `[avatar · description · ± amount]` shape used in `/feed` samples, including the `Dimmed` variant and the inline tag chip.
- **Primitives**: `bb-tx-avatar` (default + `--sm` + colored + `--uncategorized`), `bb-tx-owner-badge` (corner placement + `--letter` fallback), and `bb-tx-amount` (`--income`, `--pending`, and the combo).

Fixture rows are generated in a new `design_transaction_rows_helpers.go` so the templ stays focused on layout. Dates are seeded from `time.Now()` ("Today", "Yesterday", "4 days ago") so the relative-date column never goes stale.

## Test plan
- [x] `go vet ./...` clean
- [x] `go test ./internal/templates/components/pages/ -run Design` passes (smoke renders every section)
- [x] Visual check at `/design#transaction-rows` (1280×900) — all three variants render with category colors, tags, owner badges, pending clock, income tone

/design#transaction-rows — all 3 variants + primitives

### DIFF
--- a/internal/admin/design.go
+++ b/internal/admin/design.go
@@ -56,11 +56,6 @@ func DesignComponentHandler(sm *scs.SessionManager, tr *TemplateRenderer) http.H
 			return
 		}
 		data["Standalone"] = true
-		// Drop the standalone shell's max-w-6xl cap so the viewport
-		// toggle's Free mode (≥1320px so xl: utilities trigger) has
-		// room to grow without forcing a body horizontal scroll on
-		// reasonable browser widths.
-		data["WideMain"] = true
 		tr.RenderWithTempl(w, r, data, pages.DesignComponent(pages.DesignComponentProps{
 			Section: section,
 			Breadcrumbs: []components.Breadcrumb{
--- a/internal/admin/design.go
+++ b/internal/admin/design.go
@@ -56,6 +56,11 @@ func DesignComponentHandler(sm *scs.SessionManager, tr *TemplateRenderer) http.H
 			return
 		}
 		data["Standalone"] = true
+		// Drop the standalone shell's max-w-6xl cap so the viewport
+		// toggle's Free mode (≥1320px so xl: utilities trigger) has
+		// room to grow without forcing a body horizontal scroll on
+		// reasonable browser widths.
+		data["WideMain"] = true
 		tr.RenderWithTempl(w, r, data, pages.DesignComponent(pages.DesignComponentProps{
 			Section: section,
 			Breadcrumbs: []components.Breadcrumb{
--- a/internal/templates/components/pages/design_component.templ
+++ b/internal/templates/components/pages/design_component.templ
@@ -13,13 +13,23 @@ templ DesignComponent(p DesignComponentProps) {
 					<p class="text-sm text-base-content/50 mt-1">{ p.Section.Description }</p>
 				}
 			</div>
+			<!-- Viewport toggle. Mobile / Tablet pin the iframe to a
+			     fixed device width. Desktop fills the standalone shell's
+			     page column. Free starts at 1320px (above Tailwind
+			     xl=1280px so xl: utilities engage — e.g. TxRow's inline
+			     category picker column) and exposes a native browser
+			     resize handle on the bottom-right corner so reviewers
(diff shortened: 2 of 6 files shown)


canalesb93:
Follow-up: confirmed the picker now renders inside the standalone view's iframe at Desktop. Two fixes in `ca84197c`:

1. **Wider desktop iframe.** Bumped to `min-w-[1320px]` and wrapped the card in `overflow-x-auto`. The iframe's own `window.innerWidth` is now 1318 (vs 1102 before), so `xl:` matches and the inline category picker column renders.
2. **Seeded `window.__bbCategories`.** Inlined a 4-category fixture via `@templ.JSONScript` + a 1-line synchronous script so Alpine's `x-text="displayLabel"` resolves to the real category name instead of an empty string. Picker button is now ~90px wide showing "Groceries"/"Coffee & Cafés"/"Uncategorized" as on `/transactions`.

/design/c/transaction-rows — picker column visible after fix

canalesb93:
Second iteration in `e567aa52`:

1. **Added a new "Free" viewport toggle** alongside Mobile / Tablet / Desktop. Starts the iframe at 1320px (just above `xl=1280px` so `xl:` utilities engage) and uses native CSS `resize-x` so you can drag the bottom-right corner to any width. The wrapper keeps `mx-auto`, so the iframe stays horizontally centered as it grows. An `x-effect` clears the inline width when you switch back to a fixed mode so re-entering Free starts at the 1320px default again.
2. **Dropped the standalone shell's `max-w-6xl mx-auto`** on `/design/c/{slug}` via a new `.WideMain` branch in `base.html`. `/design` (the gallery) keeps the existing 6xl cap. No more `overflow-x-auto` crutch.

Verified at viewport 1600×900: Desktop = 1552px (fills shell), Free = 1320px centered at x=140, resize handle present, switching modes clears any user-set inline width.

/design/c/transaction-rows — Free mode (1320px, h-centered, drag handle bottom-right)

canalesb93:
Third iteration in `e460a584`:

1. **Standalone shell keeps `max-w-6xl mx-auto` by default.** The previous `WideMain` server flag was too aggressive — it dropped the cap on every visit. Now an `x-effect` on `DesignComponent` toggles `main.style.maxWidth = 'none'` only while `vp === 'free'`, with an `x-destroy` mirror that restores it on Alpine teardown. `/design` (gallery) + Mobile/Tablet/Desktop modes now look exactly like the rest of the admin.
2. **Custom right-edge grabber instead of CSS `resize: horizontal`.** A 24px primary-tinted strip with a left border + Lucide `grip-vertical` icon at vertical center sits flush against the iframe's right edge — much harder to miss than the corner triangle. `onDragStart` updates `freeWidth` by `2 × cursor-delta` so the iframe grows symmetrically around its centered axis (the cursor stays glued to the right edge as both sides expand outward). `pointer-events: none` on the iframe during drag so the embedded document doesn't swallow mousemove.

Verified at 1600×900: Desktop = 1104px wrapper inside 1152px main (default cap), Free = 1320px centered (140px each side), drag +100px → 1520px centered (40px each side), back to Desktop = all inline styles cleared.

/design/c/transaction-rows — Free mode with custom right-edge grabber

canalesb93:
Final touch in `369c8b3e`: shrank the iframe to `calc(100% - 32px)` in Free mode so there's an 8px wrapper-bg gap between the row chevrons and the 24px grabber strip. The grabber now reads as a distinct rail instead of butting flush against the rightmost row content.

/design/c/transaction-rows — grabber with left spacing

PR #1506, opened 2026-05-25
feat(admin): one-time encryption-key reveal during onboarding

canalesb93:
## Summary

Adds `/setup/save-key`, a one-time reveal of `ENCRYPTION_KEY` shown after admin creation and before any bank credentials are encrypted. At that moment the DB has no encrypted tokens to leak, so the blast radius of a copy/screenshare leak is essentially zero — the moment that changes is when a bank is connected.

Three actions cover the password-manager universe:

- **Copy** — clipboard, works everywhere.
- **Download .env line** — a one-line `.env` file the admin can drop into their secrets manager or new server.
- **Save in 1Password** — vendored [`@1password/save-button`](https://www.1password.dev/web/add-1password-button-website) web component. Inert without the extension installed, so it degrades gracefully for the ~70% of admins who use Bitwarden / Apple Passwords / pass / nothing.

After clicking "I've saved it", the timestamp lands in `app_config.setup.encryption_key_acknowledged_at` and the page redirects to `/getting-started` forever after. Recovery from that point is:

- `.env` on the server (or wherever the operator set the env var)
- `breadbox reveal-key` — new shell-only CLI subcommand that prints the resolved `ENCRYPTION_KEY` to stdout

A soft hint banner on `/getting-started` nudges admins toward the reveal page until they've acknowledged.

### `/setup/save-key` — the reveal page

<table>
  <tr><th>Desktop (1280×800)</th><th>Mobile (390×844)</th></tr>
  <tr>
    <td>save-key desktop</td>
    <td>save-key mobile</td>
  </tr>
</table>

### `/getting-started` with soft hint banner

<table>
  <tr><th>Desktop (1280×800)</th><th>Mobile (390×844)</th></tr>
  <tr>
    <td>getting-started desktop</td>
    <td>getting-started mobile</td>
  </tr>
</table>

The "Save in 1Password" button renders inert in screenshots because headless Chromium doesn't have the extension. That's the intended fallback shape.

## Design rationale

The previous design conversation landed on a few specific choices worth re-stating:

- **Why during onboarding and not later?** The DB has no encrypted credentials yet — leaking the key here is essentially harmless. After even one bank is connected, the same UI would be a hole. So we only show it in the pre-bank window, then close it permanently.
- **Why no retype/confirm-by-typing-last-N gate?** Considered, dropped. Without a hard "you really saved it" signal, retype-confirmation is theatre — and the key isn't actually lost if they skip (`.env` + `breadbox reveal-key` recover it). Replaced with a simple "I've saved it" button.
- **Why not gate provider configuration on acknowledgment?** Same logic. The acknowledgment flag is only used to suppress the page on future logins — not to lock features. Hard gates with no real signal aren't worth their UX cost.
- **Why not pull in the 1Password Go SDK (`op://` references) to avoid plaintext `.env` entirely?** It works, but bundling one vendor's SDK into the binary is the wrong shape — Bitwarden has `bws run`, Vault has `vault agent`, Doppler/Infisical have CLI wrappers. The universal pattern is `op run -- breadbox serve` (or any equivalent), which works with zero code in our binary. We can document that pattern in `breadbox-docs` as a follow-up instead.

## File layout

```
internal/appconfig/keys.go              + KeyEncryptionKeyAcknowledgedAt constant
internal/admin/setup_save_key.go        + SaveKeyHandler (GET/POST) + 1Password payload encoder
internal/admin/setup_save_key_test.go   + unit tests for the payload encoder + title helper
internal/admin/router.go                + /setup/save-key route (RequireAdmin)
internal/admin/getting_started.go       + ShouldSaveKey prop wiring
internal/templates/.../setup_save_key.templ + new page (wizard layout)
internal/templates/.../getting_started.templ + soft hint banner
static/js/admin/components/setup_save_key.js + reveal/copy/download JS (vanilla, no Alpine)
static/js/vendor/onepassword-save-button.js  + vendored from @1password/save-button 1.3.0 (~32KB)
internal/cli/reveal_key.go              + breadbox reveal-key subcommand
docs/cli-commands.md                    + reveal-key row
```

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./...` — all unit tests pass
- [x] `go test -tags=integration -p 1 ./internal/admin/...` — passes
- [x] New unit tests cover `encodeOnePasswordLogin` payload + `saveKeyItemTitle` helper
- [x] Manual end-to-end check: login → GET `/setup/save-key` renders → 1Password value attribute decodes to the expected JSON → POST stamps `setup.encryption_key_acknowledged_at` → subsequent GET redirects → `/getting-started` banner disappears
- [x] `breadbox reveal-key` prints the resolved key when set, errors when unset
- [x] Desktop + mobile screenshots attached above

## Notes

- Headless / lite builds were broken on `main` before this PR (pre-existing templ-component build-tag issues); this PR does not introduce new regressions there.
- The vendored 1Password JS bundle (~32KB) is embedded into the binary via the existing `static/embed.go` `//go:embed all:js` directive. No CDN, no phone-home.
- The reveal page uses the existing `layout.Wizard` shell. Plain vanilla JS for the page interactions (the wizard layout deliberately doesn't load Alpine).

### DIFF
--- a/docs/cli-commands.md
+++ b/docs/cli-commands.md
@@ -41,6 +41,7 @@ These commands operate on the local box (filesystem, DB, embedded migrations) �
 | `breadbox mcp` | L | MCP server over stdio; Claude Desktop spawns this per session, blocks until stdin closes. Uses a singleton `actor_type='system'` api_keys row so the audit trail attributes every stdio call consistently |
 | `breadbox mcp-stdio` | L | *deprecated — use `breadbox mcp` instead*. Hidden alias kept so existing Claude Desktop configs keep working |
 | `breadbox init` | L | First-run setup: encryption key, first login account, first API key |
+| `breadbox reveal-key` | L | Print the configured `ENCRYPTION_KEY` to stdout for recovery into a password manager. Exit 1 when no key is set |
 | `breadbox migrate [--down] [--to N]` | L | Run goose migrations against `DATABASE_URL` (local-only — there is no remote migration endpoint by design) |
 | `breadbox doctor [--skip-external] [--with-live]` | R/L | Remote mode (when a host is configured) consumes `GET /api/v1/headless/bootstrap`; local mode (no host) keeps the env/DB/provider preflight checks. `--with-live` additionally fires the agent SDK round-trip (~$0.01, local-mode only) |
 | `breadbox agent test` | L | End-to-end smoke test of the Claude Agent SDK subsystem — verifies credential is configured, sidecar binary is discoverable, and a tiny "say OK" prompt round-trips through the SDK. Cost-bounded to ~5¢. Exit 3 = no auth; exit 5 = no binary |
--- a/input.css
+++ b/input.css
@@ -956,6 +956,28 @@
     overflow-x: clip;
   }
 
+  /* ── /setup/save-key: single-line scrollable key display.
+     A 64-char hex string is wider than the wizard card on every viewport,
+     so we render it on one line with horizontal scroll inside the field
+     rather than wrapping. Keeps the action buttons below at a predictable
+     vertical position and matches the visual of password-reveal patterns
+     elsewhere (GitHub recovery codes, Auth0 client secrets, etc). */
+  .bb-key-display {
+    @apply bg-base-200/60 border border-base-300 rounded-lg
+           px-3.5 py-3 font-mono text-xs leading-relaxed
+           text-base-content/85
+           whitespace-nowrap overflow-x-auto overscroll-x-contain;
+    /* Thin, theme-aware scrollbar so it doesn't dominate the field. */
+    scrollbar-width: thin;
+  }
+  .bb-key-display::-webkit-scrollbar {
+    height: 6px;
+  }
+  .bb-key-display::-webkit-scrollbar-thumb {
+    background-color: color-mix(in oklab, var(--color-base-content) 20%, transparent);
+    border-radius: 3px;
+  }
+
   .bb-wizard-orb {
     @apply hidden;
   }
--- a/input.css
+++ b/input.css
@@ -940,10 +940,20 @@
   /* ── Wizard / Auth pages (login, setup, error) ── */
   .bb-wizard-bg {
     /* Mobile: top-aligned with generous top padding so content isn't dead-center
-       in a sea of dark background. sm+: vertically centered as before. */
+       in a sea of dark background. sm+: vertically centered as before.
+       overflow-x clip prevents off-screen children (e.g. the 1Password save
+       button's hidden info modal) from creating a horizontal scrollbar at
+       narrow widths like 320px. */
     @apply bg-base-200 min-h-screen flex items-start justify-center
-           px-4 pt-10 pb-8
+           px-4 pt-10 pb-8 overflow-x-clip
            sm:items-center sm:py-4;
+  }
+  /* Pair with .bb-wizard-bg's overflow-x: clip — without applying it to
+     <html> as well, the document root keeps the larger scrollWidth and
+     mobile browsers may still produce a horizontal scrollbar. Scoped to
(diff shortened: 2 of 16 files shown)


canalesb93:
## UI polish round

Iterated on the page in browser at four breakpoints (320 / 390 / 768 / 1280) and fixed the issues that surfaced:

- **64-char key was being truncated by the `<input>`** — replaced with a wrap-friendly monospace code block. Now fully readable when revealed (2 lines desktop, 2–3 on narrow mobile) and still masked with bullets at rest, rendered server-side so a slow JS load can't briefly leak the plaintext.
- **1Password button was fighting a shared 2-up grid** — collapsed to a single-column action stack (Copy / Download / 1Password) so the web component sits at its natural width.
- **Two competing solid-black primary buttons (Copy + Continue)** — Copy is now `btn-primary btn-soft`, Continue stays the only solid primary CTA.
- **Horizontal scroll at 320 px width** — caused by the 1Password component's hidden info modal sitting at left: -6, right: 327 in shadow DOM. Fixed with `overflow-x: clip` on `.bb-wizard-bg` + scoped `html:has(body.bb-wizard-bg)` so the document root clips too.
- **`--` instead of em-dash** in the Continue button.
- **`bg-primary/10 text-primary/70` inline code** in the "Lost it later" callout — replaced with daisy `kbd kbd-xs` so `.env` / `breadbox reveal-key` read as commands.
- **Both Copy and Download flash a 2s green check + result label** on click.

Added `internal/templates/components/pages/setup_save_key_test.go` covering the SSR masking helper and asserting the plaintext key never appears in the visible `<span data-key-text>` (only in the `data-key-value` attribute JS swaps in on reveal).

### Before → After

<table>
<tr><th>Before — desktop</th><th>After — desktop (with success feedback)</th></tr>
<tr>
<td>before desktop</td>
<td>after desktop</td>
</tr>
<tr><th>Before — mobile</th><th>After — mobile</th></tr>
<tr>
<td>before mobile</td>
<td>after mobile</td>
</tr>
</table>

### Responsive sweep (after)

<table>
<tr><th>320 px (small mobile)</th><th>768 px (tablet)</th></tr>
<tr>
<td>320px</td>
<td>768px</td>
</tr>
</table>

### Verified

- ✅ No horizontal overflow at 320 / 390 / 768 / 1280
- ✅ Reveal/hide toggle swaps key + dots and updates `aria-pressed`
- ✅ Copy puts the actual 64-char hex on the clipboard
- ✅ Download produces a `breadbox.env` file
- ✅ `go vet ./... && go test -count=1 ./...` pass

canalesb93:
## Layout tightening + drop Download + getting-started banner

Per feedback:

- **Dropped the Download .env button**. Most users go straight to a password manager; the third option was visual noise.
- **Replaced the native 1Password button with a daisy-styled proxy**. The web component sits off-screen (`position:absolute; opacity:0; pointer-events:none`) and our daisy `btn btn-primary btn-soft` forwards the click into its shadow-DOM button — same extension hook, our wrapping.
- **Two-column action row**: Copy on the left, Save in 1Password on the right.
- **Encryption key on one line, h-scrolls** inside the field. Predictable vertical position for the buttons below. Promoted to `.bb-key-display` with a thin theme-aware scrollbar.
- **Removed the soft "Save your encryption key" hint banner from `/getting-started`** along with `ShouldSaveKey` from the props and the `appconfig` check in the handler.

### Save-key — desktop (masked → revealed)

<table>
<tr><th>Masked</th><th>Revealed</th></tr>
<tr>
<td>masked</td>
<td>revealed (one-line, scrolls right)</td>
</tr>
</table>

### Save-key — mobile

<table>
<tr><th>390 px (iPhone-ish)</th><th>320 px (smallest)</th></tr>
<tr>
<td>mobile</td>
<td>320px</td>
</tr>
</table>

### Getting-started after removing the banner

getting-started without banner

PR #1517, opened 2026-05-26
fix(agents): sidecar robustness — load ESM cli.js, surface stderr, isolate HOME

canalesb93:
## Summary

Three protective fixes in the agent sidecar plus a small JSON-tag fix on `SmokeResult`. Each one resolves a distinct failure mode that's hit either on macOS dev machines or when an environment touches the bundled cli.js / `~/.claude.json` in ways the prod Linux VM never does.

## What broke

- **`Claude Code process exited with code 1`** on every run, with no further detail. The bundled `cli.js` is ESM; the sidecar extracts it to `$TMPDIR/breadbox-agent-sidecar/cli-<hash>.js` so the SDK's `fs.existsSync` check can read it, but there's no sibling `package.json` declaring `"type": "module"` — so when the SDK spawns `node <path>`, node treats it as CommonJS, refuses to load the import statements, and exits 1 instantly. Failure is silent at the operator's UI: the row just says "exited with code 1".

- **No way to debug it from the run row.** The SDK swallows the subprocess stderr by default. The Go side records `result.Err.Error()` on the row, which is the same "exited with code 1" string with no underlying cause attached.

- **`EPERM: ~/.claude.json`** on macOS dev when another Claude Code process (e.g. cmux) holds the global config file. The spawned subprocess can't write the OAuth cache / FSWatcher state and exits immediately. Doesn't reproduce on Linux prod.

- **Smoke-test result panel cells all empty** (`0 ms`, `$0.0000`, `—`). `agent.SmokeResult` had no `json:` tags, so the JSON keys came out PascalCase (`AuthMode`, `DurationMs`, …) while the templ reads snake_case (`auth_mode`, `duration_ms`, …). The `binary_path` cell was also empty because the field stashed the raw `agent.runtime_path` appconfig value, which is empty when no override is set.

## What changed

`agent/sidecar/index.ts`:
1. `resolveCliPath` now also writes a `{"type":"module"}` `package.json` next to the extracted `cli.js`.
2. The SDK `query()` call now wires `stderr: (chunk) => stderrBuf.push(chunk)`, and the catch / no-message branches append the captured stderr (truncated to ~2 KB) to the emitted error event via a new `formatStderr` helper.
3. New `configureRuntimeEnv` carves a per-run scratch dir under `$TMPDIR` and exports `HOME` + `CLAUDE_CONFIG_DIR` to it before spawning, so Claude Code stops fighting the operator's logged-in Claude Code session for the global config file.

`internal/agent/smoketest.go`:
4. `SmokeResult` gets `json:"snake_case"` tags so the diagnostics panel actually renders. `AssistantText` maps to `response` (matches the existing templ field name). `BinaryPath` is now populated via `LocateBinary(sidecarPath)` so the panel reports where the sidecar was actually picked up from (the `./bin`, user-home install, or PATH lookup).

## Why these are safe in prod

- **package.json shim** is purely additive: it only writes a file next to one we're already extracting. The extracted `cli.js` doesn't change. If node ever stopped requiring the shim, the file would just be ignored. The current Linux prod VM happens to work without it because of a kernel/node version combination, but the moment that drifts, the same symptom appears there too.
- **stderr capture** never changes the success path; it only enriches the error event payload. Existing run-detail rendering already shows `error_message` verbatim.
- **HOME isolation** is a no-op on Linux prod (the scratch dir gets written instead of the home dir, no one notices). On any host where the previous global-config write succeeded, the new per-run dir write also succeeds and is cleaned up by the OS's `$TMPDIR` lifecycle.
- **JSON tags + binary path** are pure UI plumbing; no runtime change.

## Test plan
- [ ] Existing `internal/agent/sidecar_test.go` and `smoketest_test.go` continue to pass (build/vet/test all green locally).
- [ ] Smoke test under Settings → Agents renders all six fields populated.
- [ ] Modal-fired manual run on macOS dev completes end-to-end (was previously failing with the EPERM described above).
- [ ] An intentionally-failed run surfaces the subprocess stderr inline on the run row (verified by inspecting `transcripts/agents/*.ndjson`).

### DIFF
--- a/agent/sidecar/index.ts
+++ b/agent/sidecar/index.ts
@@ -16,6 +16,7 @@ import cliAsset from "@anthropic-ai/claude-agent-sdk/cli.js" with { type: "file"
 import { existsSync, mkdirSync, writeFileSync, statSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
+import { randomBytes } from "node:crypto";
 import { JobSpecSchema, type JobSpec } from "./spec";
 import { emit, emitError } from "./events";
 
@@ -24,6 +25,12 @@ import { emit, emitError } from "./events";
 // binary, cliAsset resolves to a bunfs path that the SDK's spawn helper
 // cannot read. We materialize once per process startup, cached by mtime+
 // size so repeated cold-starts on the same binary reuse the extracted copy.
+//
+// Also writes a sibling package.json with {"type": "module"} so node
+// (the executable we pin in query()) treats the extracted .js as ESM.
+// Without it node refuses to load the SDK's import statements and the
+// spawned subprocess exits 1 with no useful event — surfacing here as
+// "Claude Code process exited with code 1".
 async function resolveCliPath(): Promise<string> {
   const dir = join(tmpdir(), "breadbox-agent-sidecar");
   mkdirSync(dir, { recursive: true });
@@ -31,6 +38,10 @@ async function resolveCliPath(): Promise<string> {
   const cached = join(dir, `cli-${bytes.length}.js`);
   if (!existsSync(cached) || statSync(cached).size !== bytes.length) {
     writeFileSync(cached, bytes);
+  }
+  const pkgPath = join(dir, "package.json");
+  if (!existsSync(pkgPath)) {
+    writeFileSync(pkgPath, '{"type":"module"}');
   }
   return cached;
 }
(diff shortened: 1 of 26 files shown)


canalesb93:
* **#1518** <a href="https://app.graphite.com/github/pr/canalesb93/breadbox/1518?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1517** <a href="https://app.graphite.com/github/pr/canalesb93/breadbox/1517?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/canalesb93/breadbox/1517?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1545, opened 2026-05-26
refactor(agents): strip agent-run row to one status signal + one body line

canalesb93:
## Summary

Rebuild `AgentRunRow` aggressively for clarity. The previous shape repeated each run's status three times (icon-tile color, status badge text, cost-column spinner), repeated time three times (`3m ago`, `32.4s`, `finished 23:46`), carried per-row trigger badges that read `cron` on ~90% of rows, and pushed the actual outcome (the report title) below a meta line of `shortID · turns · tokens` no one scans for.

The new row keeps one signal per axis:

- **Status** → just the colored icon tile (textual status badge, cost-column spinner, and `finished HH:MM` duplicate are gone).
- **Time** → `3m ago · 32s` in the title row, right-aligned. Cost is the *only* thing in the trailing column.
- **Body** → one optional line, picked in priority order: error message → top report title → skipped note. Reports were chips below the meta; promoting the top-priority report's title to the body line surfaces the actual outcome. Extras collapse into a quiet `+N` ghost badge.
- **Trigger / cap** → tooltip-only icons in the title row. `cron` (the default) is silent; only `manual`/`webhook`/`initial` and a hit cap surface as small dimmed accessories.
- **shortID, tokens, turns** → moved to the run-detail page.

The scaffold is now daisy `list` / `list-row` — the bespoke `rounded-xl … divide-y` wrapper and the hand-rolled report chips are both gone, per `.claude/rules/daisyui.md`.

`AgentRunStatusBadge` stays exported — the run-detail header and the `ShowAgent: false` rows (where the agent name is suppressed) still use it.

## Before / after

<table>
  <tr>
    <td>design sandbox — before</td>
    <td>design sandbox — after</td>
  </tr>
</table>

Live `/agents` page rendering against real data (after):

/agents page — after

Mobile (390 wide, after):

design sandbox — mobile after

## Test plan

- [x] `go build ./...` clean
- [x] `templ generate` clean (regenerates `agent_run_row_templ.go`, `agent_runs_skeleton_templ.go`, `design_sections_templ.go`)
- [x] `make css` rebuilt against the new class set
- [x] `/design/c/agent-run-rows` renders all variants (desktop + mobile)
- [x] `/agents` renders against real run data with no layout shift
- [ ] CI green

### DIFF
--- a/internal/admin/agent_run_live.go
+++ b/internal/admin/agent_run_live.go
@@ -124,14 +124,12 @@ func AgentRunLiveHandler(svc *service.Service, sm *scs.SessionManager, tr *Templ
 		}
 		payload.TranscriptHTML = buf.String()
 
-		// Render the status badge so the in_progress→success transition
-		// can update without a page reload.
-		buf.Reset()
-		if err := components.AgentRunStatusBadge(run.Status).Render(ctx, &buf); err != nil {
-			writeAdminError(w, http.StatusInternalServerError, "RENDER_ERROR", "Failed to render status badge")
-			return
-		}
-		payload.StatusBadgeHTML = buf.String()
+		// Status badge for terminal states only, inlined here so we
+		// stay on plain daisy `badge` markup. in_progress sends an
+		// empty fragment — the "live" indicator on the header meta
+		// line carries that state, and loading shouldn't sit inside
+		// a badge.
+		payload.StatusBadgeHTML = agentRunLiveStatusBadgeHTML(run.Status)
 
 		if row.DurationMs > 0 {
 			payload.DurationMs = row.DurationMs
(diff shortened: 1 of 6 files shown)


canalesb93:
Pushed [1c59549c](https://github.com/canalesb93/breadbox/pull/1545/commits/1c59549c) addressing review feedback:

- **Time moved into the title row** as a dimmed inline suffix (no more right-aligned column). Right column is cost only.
- **`AgentRunStatusBadge` deleted.** The component wasn't earning its keep — daisy `badge badge-soft badge-{tone} badge-xs` is the primitive, used directly at the row's per-agent slot, the run-detail header, and the live SSE patch. No wrapper.
- **No badge for `in_progress` anywhere** — the icon-tile spinner on the row and the "live" indicator on the detail-header meta line carry that state. Loading shouldn't sit inside a "badge".
- **Per-agent scope (ShowAgent: false)**: now leads with the relative time as the title — the icon tile already carries status, so there's no parallel text label.

### Design sandbox

design sandbox — after feedback

### Live `/agents` against real data

/agents — after feedback

### Run detail header (inline daisy `success` badge)

run detail header — after feedback

PR #1700, opened 2026-06-02
feat(dev): managed dev servers — deterministic ports, auto-cleanup, one-shot UI validation

canalesb93:
Makes running + validating Breadbox locally smooth across many worktrees and parallel agent sessions — no port fishing, no orphaned servers, one-command UI validation.

## Why

This came out of a retro on a tiny UI PR that took ~10 min: the change was 30s, the rest was wrestling the dev loop — fishing for a free port (8080–8099 saturated by parallel jobs), recovering the encryption key, fighting a locked Chrome-MCP profile, and a server that outlived the session. Root causes:

- **Port fishing.** Port logic was duplicated (the session-start hook *and* `make dev`), and it broke entirely when a worktree was created mid-session (the hook had already run against the main repo and exited early) — so the agent had no assigned port and scanned by hand.
- **Stale instances.** There was **no SessionEnd hook**. Nothing stopped a `breadbox serve` when a session closed, so they accumulated across 8081–8099. `make dev-stop` nuked the whole range (killing *active* siblings), so it wasn't safe to run casually.
- **Slow validation.** No reusable server, no checked-in screenshot script, and the Chrome DevTools MCP profile is frequently locked by a concurrent session.

## What

**One source of truth.** All port/server logic lives in `scripts/dev-lib.sh`; the Makefile, both session hooks, and the validation scripts call it. Servers are tracked in a registry under `~/.local/share/breadbox/dev-servers/`, keyed by worktree.

**No more fishing.** Port resolution is lazy + on-demand, so it works even when the session-start hook never ran. `make dev` / `make dev-watch` now auto-resolve a free port instead of hard-failing. The main checkout keeps 8080; only linked worktrees get 8081–8099.

**Auto-cleanup.** A new `SessionEnd` hook stops *this worktree's* server when the session closes (skipped on `/clear`). A reaper runs on every `SessionStart` and via `make dev-reap`, killing servers whose pid is dead or whose worktree was removed. `make dev-stop` is now worktree-safe; the old range-nuke is `make dev-stop-all`.

**One-command validation.** `make dev-shot ARGS="/transactions --mobile"` rebuilds, (re)starts a tracked background server, and screenshots via puppeteer + system Chrome with a fresh profile — never colliding with a locked Chrome-MCP session.

### Command surface (all match the `make dev*` sandbox exclusion)

```sh
make dev-bg                                   # start/reuse a background server; prints URL
make dev-shot ARGS="/transactions --mobile"   # rebuild + restart + screenshot → JPEG paths
make dev-ps                                    # list every managed server
make dev-stop                                  # stop THIS worktree's server (safe for siblings)
make dev-reap                                  # kill orphans (dead pid / removed worktree)
make dev-stop-all                              # blunt: kill everything on 8080-8099
```

### Design choices

- **On-demand, not auto-boot.** Booting a server at every session start would *add* servers (most sessions never touch UI) and worsen the stale problem. `make dev-bg` is instant-and-reusable instead.
- **Plain `breadbox serve` (not `air`) for the managed server** — one process = a trivial, reliable lifecycle. `dev-server restart` rebuilds. `make dev-watch` still exists for interactive hot-reload.
- Resilient to the stale-`internal/db` gotcha: if `go build` fails, it forces `make sqlc && make templ` and retries once.

## Review

Self-reviewed at `/code-review high` (7 finder angles); fixed the confirmed issues **before** this commit:

- `dev-port` no longer writes registry entries — the `make dev` foreground path never cleared them, leaking stale `RESERVED` claims. The registry is now owned solely by managed background servers (whose port-bind already protects the port).
- `bb_load_env` tolerates `set -u`/`set -e` while sourcing `.local.env` (an unset-var reference no longer aborts the caller).
- `bb_stop_port` guards `kill` against `set -e` and drops a redundant lsof-listener block — the `exec`-based launch makes the recorded pid the real listener, so the block only added a sibling-kill TOCTOU race.
- `session-end` parses the hook JSON without a hard `python3` dependency (sed fallback), so `/clear` is honored on stock macOS, and skips killing a server another session owns.
- Reservations are keyed by the canonical git worktree root; stale `PENDING` boots are reaped by age.
- `disown` fixed (jobspec, not pid); faster `lsof -nP -sTCP:LISTEN`; dead `.gitignore`/`server.pid` removed.

## Test plan

- [x] `bash -n` + `node --check` on every script/hook
- [x] Unit test of the lib (claim/idempotency/lookup/reap/meta-write/stop)
- [x] E2E: boot → recorded pid **==** port listener → idempotent reuse → screenshot → stop frees the port + empties the registry
- [x] SessionEnd hook: `reason=clear` leaves the server up; `reason=other` stops it
- [ ] Reviewer: confirm the SessionEnd hook addition doesn't surprise in-flight sessions (only new sessions load it; existing ~20 won't run it until restarted)

### DIFF
--- a/.claude/hooks/session-end.sh
+++ b/.claude/hooks/session-end.sh
@@ -31,8 +31,14 @@ except Exception: pass" 2>/dev/null || true)"
 }
 CWD="$(bb_json_field cwd)"
 REASON="$(bb_json_field reason)"
-SESSION="$(bb_json_field session_id)"
 [ -n "$CWD" ] || CWD="$PWD"
+# Registry entries are tagged with the ENV session id (bb_session_id). Read the
+# session from the SAME source here so a same-session comparison is identical by
+# construction; only fall back to the stdin JSON id if the env var is absent.
+# (Using a different source than the writer could make the owner check think a
+# sibling owns the server and skip cleanup — a leak. Same source avoids that.)
+SESSION="${CLAUDE_CODE_SESSION_ID:-${CLAUDE_SESSION_ID:-}}"
+[ -n "$SESSION" ] || SESSION="$(bb_json_field session_id)"
 
 # Locate the lib relative to the active worktree if the project-dir copy is absent.
 LIB="$PROJECT_DIR/scripts/dev-lib.sh"
--- a/.claude/hooks/session-end.sh
+++ b/.claude/hooks/session-end.sh
@@ -1,0 +1,62 @@
+#!/bin/bash
+# SessionEnd hook — stop the managed dev server this session's worktree started,
+# and reap any orphaned servers left by previously-closed sessions.
+#
+# This is the cleanup that was missing: nothing used to stop a `breadbox serve`
+# when a session closed, so stale instances accumulated across 8081-8099. We
+# only stop the CURRENT worktree's managed server (tracked in the dev-server
+# registry) — sibling worktrees' servers are left alone.
+#
+# Skipped on reason=clear (a /clear keeps the session alive; killing the dev
+# server then would be hostile). Reap still runs (cheap, only touches dead
+# entries).
+
+PROJECT_DIR="${CLAUDE_PROJECT_DIR:-$(pwd)}"
+
+# Hook input arrives as JSON on stdin (cwd, reason, session_id). Parse without a
+# hard python3 dependency (stock macOS often lacks it): try python3, then fall
+# back to a sed string-field extractor.
+INPUT="$(cat 2>/dev/null || true)"
+bb_json_field() { # <key> — reads $INPUT
+  local key="$1" val=""
+  if [ -n "$INPUT" ] && command -v python3 >/dev/null 2>&1; then
+    val="$(printf '%s' "$INPUT" | python3 -c "import sys,json
+try: print(json.load(sys.stdin).get('$key',''))
+except Exception: pass" 2>/dev/null || true)"
+  fi
+  if [ -z "$val" ] && [ -n "$INPUT" ]; then
+    val="$(printf '%s' "$INPUT" | sed -n "s/.*\"$key\"[[:space:]]*:[[:space:]]*\"\([^\"]*\)\".*/\1/p" | head -1)"
+  fi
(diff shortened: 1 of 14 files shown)


canalesb93:
Pushed `439336b9` — fixes from a follow-up `/code-review high` of this branch.

**Confirmed bugs fixed**

1. **session-end ownership guard was dead code.** Entries were tagged with `$CLAUDE_SESSION_ID`, but the real env var is `CLAUDE_CODE_SESSION_ID`, so the `session` field was always empty and session-end always killed the server — defeating the "leave a sibling session's server running" guard. Now uses a single `bb_session_id()` helper (accepts both names), RESERVED claims carry the session field too, and session-end reads the session from the **same** env source as the writer (stdin JSON only as fallback) so a same-session comparison matches by construction.
2. **`make dev PORT=N` / `make dev-watch PORT=N` was silently ignored in worktree sessions** — `dev-port` preferred the session-exported `SERVER_PORT` over the explicit CLI `PORT`. The Makefile now forwards `BB_PORT_EXPLICIT=1` (via `$(origin PORT)`) and `dev-port` honors an explicit override ahead of the ambient `SERVER_PORT`.
3. **`make dev-stop-all` could orphan untrackable processes** — it sent only SIGTERM then wiped the whole registry. Replaced with `dev-server stop-all` → `bb_stop_all`: SIGTERM → wait → SIGKILL survivors, then drop only registry entries whose process is actually gone (a survivor stays tracked).

**Latent fix:** `bb_load_env` now restores the caller's `set -a` state instead of unconditionally clearing it.

**Verification (all green):** dev-port precedence (explicit / ambient / 8080), Makefile `$(origin)` wiring, session-tagged registry entries, session-end guard (sibling keeps · owner stops · `/clear` keeps), and `bb_stop_all` keep-live / drop-dead.

The other review findings (eager RESERVED reservations exhausting the 8081–8099 range; the `make dev` foreground-vs-background port race; PENDING age-reap during a >10 min build; duplicated port-preference logic) are real but lower-severity / more design-sensitive — left for a follow-up rather than folded in here.

PR #1798, opened 2026-06-08
design(sandbox): Workflows DNA tab + /accounts redesign prototype

canalesb93:
Adds a **Workflows DNA** group to the `/design` sandbox (bottom of the rail) that extracts the top-level patterns making `/workflows` + `/workflows/runs` (and their drawers) our cleanest surfaces, so they can be reapplied elsewhere — plus a worked `/accounts` redesign prototype applying them.

## Sections
- **The DNA** — a seven-principle manifesto (consistent chrome, status-in-one-tile, one body line by priority, minimal icon clusters, edit-in-a-slide-over, generous spacing, identity via avatar).
- **Page scaffold** — PageHeader + TabBar + box-variant status filter.
- **Gallery card** / **Run list row** — reuse the live `SectionWorkflowPresetCard` / `SectionAgentRunRows`.
- **Side drawer** — a workflows-flavored slide-over.
- **Applied — /accounts redesign (prototype)** — the table reimagined as the list-row surface, grouped by connection with light group headers and a quick-manage drawer. Sample data only; not wired to live accounts.

## Notes
- New group registered in `DesignSections()` / `DesignSectionGroups()`; also fixed a latent gap where `onboarding` was missing from `design_gallery.js`'s open-state map.
- Sandbox-only; no production surface changes.

/accounts redesign prototype in the sandbox

## Test plan
- [x] `go build ./...`, `go test ./internal/templates/...`
- [x] Rendered at `/design` (desktop + mobile); drawer opens via `$store.drawers`

PR #1799, opened 2026-06-08
feat(accounts): regroup /accounts by connection (Workflows DNA)

canalesb93:
Applies the approved `/accounts` redesign — the dense 5-column sortable table becomes the Workflows-DNA list-row surface, **grouped by connection**.

## What changed
- **Grouped by connection.** Each institution is a quiet labelled group (name · count · subtotal — the `/transactions` day-group idiom) carrying the connection health (one status per connection: a `Reauth needed` / `Disconnected` badge + a recovery link to `/connections`) above its account list-rows. Orphan accounts (connection deleted → `SET NULL`) bucket last.
- **List-rows replace the table.** A type tile, the account name (+ `Linked`/`Excluded` badges + owner avatar), a quiet `····mask · Type` line, and the balance (red for liabilities). Status is not repeated per-row — it lives on the group header. Rows stack cleanly on mobile (no column-hiding).
- **Member filter is server-side** (`?user=<short_id>`), so every group subtotal re-scopes to the selected member. Unknown ids fall back to "all".
- **Dropped** the client-side member chip strip, the search box, and clickable-header sorting — with a household's handful of accounts, grouping carries it. `accounts.js` shrinks to just the exclude/include action.
- `GroupAccountsByConnection` is a pure, unit-tested function; the loading skeleton is updated to match.

The full `/accounts/{id}` detail page (with transactions) is unchanged and remains the click-through target. A quick-manage **drawer** is a planned follow-up (prototyped in the `/design` → Workflows DNA tab, #1798).

## After
/accounts redesign — desktop

Mobile:
/accounts redesign — mobile

## Test plan
- [x] `go build ./...`, `go vet ./...`, `go test ./...` (unit) — green
- [x] New unit test `TestGroupAccountsByConnection` (ordering, subtotals, status-on-header, orphan bucket)
- [x] Verified live with real dev data: groups + subtotals + owner avatars + liabilities render; `?user=` filter + invalid-fallback return 200; exclude/include still works
- [ ] Reviewer: confirm scoping for a non-editor member (already query-scoped) reads right

### DIFF
--- a/internal/admin/accounts.go
+++ b/internal/admin/accounts.go
@@ -70,36 +70,68 @@ func AccountsListPageHandler(a *app.App, svc *service.Service, sm *scs.SessionMa
 			userAvatarVersionByShort[u.ShortID] = usersAvatarVersion(u.UpdatedAt)
 		}
 
-		// Build typed rows + running totals.
+		// Build typed rows.
 		var rows []pages.AccountsListRow
-		var totalAssets, totalLiabilities float64
-		var hasAnyBalance bool
-
-		appendRow := func(row pages.AccountsListRow) {
-			if row.HasBalance {
-				hasAnyBalance = true
-				if row.IsLiability {
-					totalLiabilities += math.Abs(row.BalanceFloat)
-				} else {
-					totalAssets += row.BalanceFloat
-				}
-			}
-			rows = append(rows, row)
-		}
-
 		if useUser {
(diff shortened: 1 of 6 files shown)


canalesb93:
Folded into the design-system sprint #1805 (same /accounts redesign commit, now the worked example on the sprint branch).

PR #1843, opened 2026-06-13
feat(mcp): restore merchant_summary tool

canalesb93:
Follow-up to #1841. Restores the `merchant_summary` MCP tool.

While auditing the Routine Reviewer's prompts I found `merchant_summary` referenced as a tool in **six** agent prompts and in `docs/mcp-tools-reference.md`, but it was **not registered** in the MCP server — agents told to "use merchant_summary" hit an unknown tool and improvised. The building blocks already existed:

- Service method `GetMerchantSummary` (`internal/service/transaction_summary.go`) — production-grade, fully parameterized.
- REST endpoint `GET /transactions/merchants` already exposes it.
- `docs/mcp-tools-reference.md` already documents it, and the admin tool-label map already lists it.

Only the MCP wiring was missing. This re-registers it.

## What it does

Per-**merchant** aggregation over a window (default last 90 days) — the merchant-grouped companion to `transaction_summary` (which groups by category/time, not merchant). Each row: `merchant`, `transaction_count`, `total_amount`, `avg_amount`, `first_date`, `last_date`, `iso_currency_code` (never summed across currencies; capped at 500).

Filters: `min_count` (2 = recurring, 3+ = subscriptions), `spending_only`, `category_slug`, `account_id`, `user_id`, `min_amount`/`max_amount`, `search` (+ `search_mode` incl. fuzzy for bank-feed name variants), `exclude_search`.

This makes the six prompts that already reference it correct, no prompt edits needed.

## Tests
- New MCP handler test: merchants array + totals shape, per-row keys, and `min_count` exclusion.
- `go build` / `go vet` clean; full unit suite + `mcp`/`admin` integration green.

## Stack
Stacked on #1841 (`agents/rule-query-and-agent-shell`) since both touch `server.go` / `server_test.go` / `tools.go` / the docs reference — basing here keeps this diff to just the merchant change. Once #1841 squash-merges, GitHub retargets this to `main` cleanly.


canalesb93:
Closing — the decision changed: we want merchant_summary **removed and scrubbed**, not restored. A better merchant-data offering is planned for the future, so the phantom references are being cleaned up instead (separate PR off main). See follow-up.

PR #1874, opened 2026-06-16
feat(rules): transaction metadata in conditions and actions

canalesb93:
## What

Makes the free-form transaction **metadata** blob a first-class read/write surface for the rule engine, usable by agents (MCP) and the REST API.

### Conditions — read custom fields
New dotted leaf field `metadata.<key>` reads any key from a transaction's metadata blob:
```json
{ "field": "metadata.tax_deductible", "op": "eq", "value": true }
{ "field": "metadata.reimburse_cents", "op": "gte", "value": 10000 }
{ "field": "metadata.project_code", "op": "exists" }
```
- Operators: `eq` `neq` `contains` `not_contains` `matches` `in` `gt` `gte` `lt` `lte` `exists` `not_exists`.
- `eq`/`neq` pick numeric vs bool vs string comparison from the **expected value's type**.
- An **absent key matches only `not_exists`** — so a `neq`/`not_contains` rule can't silently match every transaction that simply lacks the key.
- Keys are **case-sensitive** (JSONB keys), unlike tag slugs.

### Actions — write custom fields
```json
{ "type": "set_metadata", "metadata_key": "tax_deductible", "metadata_value": true }
{ "type": "remove_metadata", "metadata_key": "needs_receipt" }
```
- `set_metadata` upserts one key (value is any JSON type); `remove_metadata` deletes one key. Both repeat freely.
- **Last-writer-wins** per key; **net-diff** against each other in a pass (set-then-remove cancels); **chain** — a `set_metadata` write is visible to a later-stage rule's `metadata.<key>` condition.
- Same key (≤128 char) / value (≤4 KiB) caps as the existing scoped metadata ops.

### Coverage
Sync apply (single merged `UPDATE`), `preview_rule`, single- and all-rule retroactive apply, and `find_matching_rules`. Like `assign_series`, metadata writes emit **no dedicated timeline annotation** (the rule's `hit_count` records the firing); retroactive apply still writes the `rule_applied` audit row.

### Surfaces
- **MCP** `create_/update_/batch_create` rule tools — action maps switched to `map[string]any` so `metadata_value` keeps its JSON type. (Side benefit: this also wires `assign_series` through MCP, which was string-only before.)
- **REST** unchanged — conditions/actions are freeform JSON, no new routes (no OpenAPI change).
- Docs: `docs/rule-dsl.md` + `docs/mcp-tools-reference.md`.

## Not in this PR
The **admin rule-form UI editor** doesn't yet expose metadata conditions/actions — its single-value editor can't model a dynamic key + typed value cleanly. This matches the existing precedent that `assign_series` is API/MCP-only in the form. Coming as a stacked follow-up.

## Testing
- `go build ./...`, `go vet ./...` — clean.
- New unit tests: `internal/sync/rule_resolver_metadata_test.go`, `internal/service/rules_metadata_test.go` (conditions, actions, net-diff, chaining, validation).
- New integration tests: `internal/service/rules_metadata_integration_test.go` (retroactive set/remove + metadata-condition match). Full integration sweep of `service`/`sync`/`db`/`mcp` green.

### DIFF
--- a/docs/mcp-tools-reference.md
+++ b/docs/mcp-tools-reference.md
@@ -302,20 +302,20 @@ The tools below are thin API skins over the DSL — the DSL doc is the source of
 
 ### create_transaction_rule (Write)
 
-Create a rule that fires during sync. Actions compose (`set_category` + `add_tag` + `add_comment` in a single rule are all valid).
+Create a rule that fires during sync. Actions compose (`set_category` + `add_tag` + `set_metadata` in a single rule are all valid).
 
 | Parameter | Type | Description |
 |-----------|------|-------------|
 | `name` | string | Human-readable rule name |
-| `conditions` | object | Condition tree. Omit or `{}` for match-all. Supports `and` / `or` / `not` nesting up to depth 10. |
-| `actions` | array | Typed actions: `set_category`, `add_tag`, `remove_tag`, `add_comment`. Either this or `category_slug` is required. |
+| `conditions` | object | Condition tree. Omit or `{}` for match-all. Supports `and` / `or` / `not` nesting up to depth 10. Leaf fields include `metadata.<key>` to read a key from the free-form metadata blob (ops: `eq`/`neq`/`contains`/`not_contains`/`matches`/`in`/`gt`/`gte`/`lt`/`lte`/`exists`/`not_exists`; an absent key matches only `not_exists`). |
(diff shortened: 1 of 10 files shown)


canalesb93:
Pushed a fix from code review: `set_metadata` followed by `remove_metadata` of the **same pre-existing key** now correctly deletes it (last-writer-wins), instead of cancelling the set and silently reverting to the original DB value. Applied across all three paths — sync resolver, single-rule retroactive, and bulk `ApplyAllRulesRetroactively` — plus unit + integration regression tests. The sync path also now deep-copies the metadata map so chaining mutations don't leak into the caller's blob.

PR #1876, opened 2026-06-16
feat(rules-ui): metadata conditions + actions in the rule form editor

canalesb93:
Stacked on #1874 (backend). The visual rule builder gains first-class **metadata** support so humans can author the new `metadata.<key>` conditions and `set_metadata` / `remove_metadata` actions in the form — no dropping to the JSON editor (and no longer API/MCP-only).

Rule form with metadata conditions and actions

## Conditions
- New **"Metadata field"** option in the field dropdown reveals a **key** input and a dedicated **operator** select: `equals` `not equals` `contains` `not contains` `regex` `in list` `> ≥ < ≤` `exists` `does not exist`.
- The value input **hides for the presence ops** (`exists` / `does not exist`) and comma-splits for `in list`.
- The dotted `metadata.<key>` field is assembled at submit time and split back into field + key when loading a rule for **edit**.

## Actions
- **Set metadata** — key + a `text` / `number` / `true/false` **type selector** + value. The value is coerced to the chosen JSON type on submit (`"2026"` → `2026`, `"true"` → `true`), so the stored blob is correctly typed.
- **Remove metadata** — key only.
- Both repeat freely; a same-key set+remove raises the existing combination warning.

## Why a special case
The builder's `{field, op, value}` condition and `{field, value}` action models have no slot for a dynamic **key** or a **typed value** — metadata rows carry extra `key` / `valueType` fields, with dedicated serialize/parse paths.

## Validation (real browser)
- Forward serialization captured from `submitRule()`: `metadata.tax_deductible eq "true"`, `metadata.project_code exists` (no value), `metadata.tags_csv in ["a","b","c"]`; actions with typed values (string / number `2026` / boolean `true`) + `remove_metadata`.
- Create → reload `/rules/{id}/edit` → state re-populates correctly (`field=metadata` + key + op + value; action `valueType` recovered as number/boolean/text) → delete. Backend accepted the POST (201) and DELETE (204).
- `go build`, `go vet`, and the pages/scripts-lint tests pass.

### DIFF
--- a/internal/templates/components/pages/rule_form.templ
+++ b/internal/templates/components/pages/rule_form.templ
@@ -193,7 +193,15 @@ templ RuleForm(p RuleFormProps) {
 										<option value="user_name">Family member</option>
 										<option value="provider">Provider</option>
 									</optgroup>
+									<optgroup label="Metadata">
+										<option value="metadata">Metadata field</option>
+									</optgroup>
 								</select>
+								<!-- Metadata key — only for the metadata field. The dotted
+								     field metadata.<key> is assembled at submit time. Fixed width
+								     + shrink-0 (matching the numeric/bool controls) so it doesn't
+								     eat the row; flex-wrap drops operator+value to row 2 on mobile. -->
+								<input type="text" x-model="cond.key" x-show="cond.field === 'metadata'" @input="syncToJson()" class="input input-xs bg-base-100 w-32 shrink-0" placeholder="key"/>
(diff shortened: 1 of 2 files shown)


canalesb93:
Pushed a fix from code review: a metadata `Set metadata` action with type **number** and a non-numeric value now surfaces an inline error and blocks save, instead of silently storing the raw string (which contradicted the chosen type). Rebased onto the updated backend branch.

